Issue #15, opened 2021-05-16
Queue backend: Support Amazon SQS


adelowo:

oneEyedSunday:
Hello Lanre, some good stuff here.

Re AWS, gokit had a PR planning for SQS support, it got tied down in reviews.
see [here](https://github.com/go-kit/kit/pull/1018) and at the last count, there are two or so competing ways of defining the interfaces.

I want to have a go at this issue, with that PR as a base.

oneEyedSunday:
Hello Lanre, some good stuff here.

Re AWS, gokit had a PR planning for SQS support, it got tied down in reviews.
see [here](https://github.com/go-kit/kit/pull/1018) and at the last count, there are two or so competing ways of defining the interfaces.

I want to have a go at this issue, with that PR as a base.

adelowo:
Hi @oneEyedSunday sure, happy to have you take a go at the issue. 

Namaste

PR #23, opened 2021-06-04
Support AWS SQS as queue backend

oneEyedSunday:
Base support for SQS, along with some configuration options.

We delete the messages after they have been passed to our internal channel.

### DIFF
--- a/queue/sqs/client.go
+++ b/queue/sqs/client.go
@@ -1,0 +1,152 @@
+package sqs
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+
+	"github.com/hookcamp/hookcamp"
+	"github.com/hookcamp/hookcamp/config"
+	"github.com/hookcamp/hookcamp/queue"
+	"github.com/hookcamp/hookcamp/util"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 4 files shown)


oneEyedSunday:
Im unsure how the queues will be consumed by the application.
It would seem wasteful to have to delete messages individually.

SQS sdks usually expose a batchDelete API (that doesn't guarantee all message handles will be deleted), so that's a more economical option.

oneEyedSunday:
For #15

adelowo:
Apologies for the delayed response @oneEyedSunday . It looks good to me. If we fix the linting issues, i can merge it

queue/sqs/client.go
```diff
@@ -0,0 +1,152 @@
...
+
+			continue
+		}
+
+		channel <- queue.Message {
+			Err: nil,
+			Data: m,
+		}
+
+		c.deleteMessage(msg.ReceiptHandle)
```
adelowo, line 132: I think we can get rid of this `deleteMessage` now. We need to verify  that the messages have been processed.

We can get this out for now

adelowo (changes requested):

queue/sqs/client.go
oneEyedSunday: ok, I'd get rid of it.

adelowo (approved):

queue/sqs/client.go
adelowo: Thanks

PR #224, opened 2021-11-19
implement force resend

danvixent:


jirevwe (approved):

server/route.go
```diff
@@ -250,6 +251,7 @@ func buildRoutes(app *applicationHandler) http.Handler {
 			eventDeliveryRouter.Use(requirePermission(auth.RoleUIAdmin))
 
 			eventDeliveryRouter.With(pagination).Get("/", app.GetEventDeliveriesPaged)
+			eventDeliveryRouter.Post("/forceresend", app.ForceResendEventDeliveries)
 			eventDeliveryRouter.Post("/batchretry", app.BatchRetryEventDelivery)
```
jirevwe, line 255: These are not RESTful at all.

jirevwe (approved):

server/route.go
danvixent: :) we can fix those much later

PR #232, opened 2021-11-22
use kelseyhightower/envconfig to load config from env

danvixent:


config/config.go
```diff
@@ -5,67 +5,79 @@ import (
...
 
 type QueueConfiguration struct {
-	Type  QueueProvider `json:"type"`
-	Redis struct {
-		DSN string `json:"dsn"`
-	} `json:"redis"`
+	Type  QueueProvider           `json:"type" envconfig:"CONVOY_QUEUE_PROVIDER"`
+	Redis RedisQueueConfiguration `json:"redis"`
+}
+
```
danvixent, line 42: @Subomi we should probably change this to :
`
type QueueConfiguration struct {
	Type  QueueProvider           `json:"type" envconfig:"CONVOY_QUEUE_PROVIDER"`
	Dsn string `json:"dsn"`
       // other details, we can switch based on the value of Type and set it up accordingly 
}
`
I suggest the same for StrategyConfiguration

Subomi (approved):

PR #241, opened 2021-11-24
Update Changelog

Subomi:


oneEyedSunday:
Hello @Subomi  can subsequent changelogs have links to the merge commit? That way its easier to follow when a feature was added.

Subomi:
Hmm, noted. Thanks @oneEyedSunday

PR #261, opened 2021-12-09
add modd.conf for live reload

danvixent:
When convoy is run using the [modd](https://github.com/cortesi/modd) tool, any changes to .go files or the convoy.json file will cause the project to be recompiled and run again. modd is configured with the modd.conf file.

Install the modd tool above and run convoy with:

```
$ modd
```


Subomi:
@danvixent I don't think I'll want us to add a new tool to the core to achieve this yet. Maybe later in the future.

PR #264, opened 2021-12-10
implement support for new relic

### LINKED ISSUE #258: Implement Support for New Relic
Add support for structured logging and instrumenting http requests via new relic

Dotunj:
This adds support for instrumenting http requests via new relic. This resolves #258 

Also added support for a logger interface and a corresponding implementation that simply writes to standard output. This also provides support for optional logging of http requests.


server/middleware.go
```diff
@@ -455,6 +482,139 @@ func pagination(next http.Handler) http.Handler {
...
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+			cfg, err := config.Get()
+			if err != nil {
+				return
+			}
+
+			ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
+
+			if isLogHttpRequestEnabled(cfg.Logger) && cfg.Logger.Type != "" {
```
Subomi, line 496: I don't think we need to check if the Logger type is not nil here. The system should default to console logger where the user don't specify a different logger.

logger/console.go
```diff
@@ -0,0 +1,58 @@
+package logger
```
Subomi, line 1: Let's write test for this guy.

cmd/main.go
```diff
@@ -98,6 +102,20 @@ func main() {
 				}
 			}
 
+			if cfg.Logger.Type != "" {
+				lo, err = logger.NewLogger(cfg.Logger)
```
Subomi, line 106: Where a logger is not provided, the system should default to console logger with log level -- info.

config/config.go
```diff
@@ -84,6 +84,22 @@ type GroupConfig struct {
...
 
+type LoggerConfiguration struct {
+	Type  LoggerProvider `json:"type"`
+	Level string         `json:"level"`
+	LogHttpRequest bool  `json:"log_http_request"`
+}
+
+type TracerConfiguration struct {
+	Type     TracerProvider `json:"type"`
+	NewRelic struct {
```
Subomi, line 95: NewRelic configuration should be standalone -- `NewRelicConfig`

tracer/tracer.go
```diff
@@ -0,0 +1,36 @@
+package tracer
```
Subomi, line 1: I'm not sure we should be doing tracing yet. I'd prefer us to figure out logs & metrics first.

Subomi (changes requested):
@Dotunj FYI boss.

tracer/tracer.go
Dotunj: If we're not doing tracing yet, then there might be no need for New Relic at the moment.

Subomi: I don't think so -- Error tracking, Instrumentation.

config/config.go
```diff
@@ -84,6 +84,22 @@ type GroupConfig struct {
 	DisableEndpoint bool                   `json:"disable_endpoint"`
 }
 
+type LoggerConfiguration struct {
+	Type  LoggerProvider `json:"type"`
+	Level string         `json:"level"`
+	LogHttpRequest bool  `json:"log_http_request"`
```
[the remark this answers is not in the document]
Subomi: We can instead tie the http requests to a particular log level vs. creating an new config level.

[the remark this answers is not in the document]
Dotunj: what log level are we going with for this?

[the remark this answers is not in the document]
Subomi: Info.

Subomi (approved):

danvixent (approved):

jirevwe (approved):

PR #268, opened 2021-12-15
Separate server from worker nodes

ogbanugot:
I have two commands to create a master node and a worker node. Leader election acquisition with Consul sessions included.


oneEyedSunday:
I'm going to ask Ogban, whats the end game here, a stop-gap to k8s?

ogbanugot:
> I'm going to ask Ogban, whats the end game here, a stop-gap to k8s?

In a way, but the primary objective is that we can horizontally scale worker nodes, create clusters of workers etc. Together with k8s, the deployment of server and worker nodes can be done separately and scaled separately too.

Subomi:
For context, this pr is going to be a long-running effort, and we're trying to get a proper sense of it. So we this approach: 
1. We are creating a branch `distributed-convoy`
2. All PRs for this effort will go there. 
3. The branch will be kept up to date with `master`.
4. The branch will be kept in a runnable state. 
5. We will continue to iterate in that branch till we are confident to release it.

PR #293, opened 2022-01-04
adds an example config file

Dotunj:
This PR adds a convoy.json.example file

### DIFF
--- a/convoy.json.example
+++ b/convoy.json.example
@@ -1,0 +1,80 @@
+{
+    "database": {
+        "environment": "dev",
+        "dsn": "mongodb://root:rootpassword@localhost:27017/convoy?authSource=admin"
+    },
+    "sentry": {
+        "dsn": "<insert-sentry-dsn>"
+    },
+    "queue": {
+        "type": "redis",
+        "redis": {
+            "dsn": "redis://localhost:8379"
+        }
+    },
+    "logger": {
+        "type": "console",
+        "server_log": {
+            "level": "info"
+        }
+    },
+    "tracer": {
+        "type": "new_relic"
+    },
+    "new_relic": {
+        "license_key": "<new-relic-license-key>",
+        "app_name": "convoy",
+        "config_enabled": true,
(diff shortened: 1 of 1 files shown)


jirevwe (approved):

convoy.json.example
```diff
@@ -0,0 +1,80 @@
+{
+  "database": {
+    "environment": "dev",
```
Subomi, line 3: @Dotunj `environment` is a top-level field not nested in `database`.

Subomi (changes requested):

convoy.json.example
```diff
@@ -0,0 +1,80 @@
...
+    "dsn": "mongodb://root:rootpassword@localhost:27017/convoy?authSource=admin"
```
jirevwe, line 4: The connection will fail in the docker network because `localhost` will not resolve to the actual `mongodb` container 
```suggestion
    "dsn": "mongodb://root:rootpassword@mongodb:27017/convoy?authSource=admin"
```

convoy.json.example
```diff
@@ -0,0 +1,80 @@
...
+  },
+  "sentry": {
+    "dsn": "<insert-sentry-dsn>"
+  },
+  "queue": {
+    "type": "redis",
+    "redis": {
+      "dsn": "redis://localhost:8379"
```
jirevwe, line 12: This connection will also fail in the docker network because `localhost` will not resolve to the actual redis server container
```suggestion
      "dsn": "redis://redis_server:8379"
```

jirevwe (changes requested):

convoy.json.example
Subomi: Exactly, why I removed it. Formerly, as well we had `convoy-docker.json` for running convoy in a container. 

I suggest we just rename it to `<insert-mongodb-dsn>` same for redis. @Dotunj

Subomi (approved):

jirevwe (approved):

PR #298, opened 2022-01-05
feat: add delete group api

### LINKED ISSUE #291: Enable Delete Group over API.
Currently, it is not possible to delete a group over the API. This should be possible.

danvixent:
This PR adds an API for deleting a group and its associated entities, fixes #291


datastore/group.go
```diff
@@ -108,3 +111,59 @@ func (db *groupRepo) FetchGroupByID(ctx context.Context,
...
+
+	groupFilter := bson.M{"uid": uid}
+	_, err := db.inner.UpdateOne(ctx, groupFilter, update)
+	if err != nil {
+		return err
+	}
+
+	appFilter := bson.M{"group_id": uid}
+
+	appCollection := db.innerDB.Collection(AppCollections)
```
Subomi, line 139: @danvixent preferably, let's create `Delete` methods for applications & events and used their methods respectively vs. writing it here.

Subomi (changes requested):

datastore/group.go
danvixent: done, but watch out for a TODO I added

Subomi: Seen. Instead, we should wrap the entire thing in a transaction. But apparently, the current design doesn't expose the datastore to the handlers to start a transaction so another PR.

Subomi (approved):

PR #300, opened 2022-01-06
fixed eventdelivery filtering bug

### LINKED ISSUE #301: EventDelivery Filtering Bug
Filtering event deliveries by status was changed to filtering by an array of status instead of a single status. But the PR that fixed it introduced a bug which was passing `[""]` as valid status to filter by.

Subomi:
Resolves #301


jirevwe:
Can you add more context on what was fixed in this PR? I don't understand the bug. Maybe with an issue?

Subomi:
> Can you add more context on what was fixed in this PR? I don't understand the bug. Maybe with an issue?

I added the issue #301

jirevwe (approved):

Issue #290, opened 2021-12-31
Perform validation checks while creating groups


Subomi:
Currently, creating groups over the API does no validation check over the payload.

PR #304, opened 2022-01-06
fix: add request validation for creating groups

Dotunj:
This adds additional request validation to the endpoint for creating a group. Addresses #290

### DIFF
--- a/server/group.go
+++ b/server/group.go
@@ -2,11 +2,14 @@ package server
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
 
 	"github.com/frain-dev/convoy"
 	"github.com/frain-dev/convoy/auth"
+	"github.com/frain-dev/convoy/config"
+	"github.com/frain-dev/convoy/config/algo"
 	"github.com/frain-dev/convoy/server/models"
 	"github.com/frain-dev/convoy/util"
 	"github.com/frain-dev/convoy/worker/task"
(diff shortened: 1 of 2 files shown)


jirevwe (approved):

server/group.go
```diff
@@ -58,6 +61,41 @@ func (a *applicationHandler) CreateGroup(w http.ResponseWriter, r *http.Request)
...
+
+	if util.IsStringEmpty(string(newGroup.Config.Signature.Header)) {
+		_ = render.Render(w, r, newErrorResponse("please provide a valid signature header", http.StatusBadRequest))
+		return
+	}
+
+	if util.IsStringEmpty(string(newGroup.Config.Signature.Hash)) {
+		_ = render.Render(w, r, newErrorResponse("please provide a valid hash", http.StatusBadRequest))
+		return
+	}
```
danvixent, line 92: is this necessary since we're still going to check algo.M for it?

server/group.go
```diff
@@ -58,6 +61,41 @@ func (a *applicationHandler) CreateGroup(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	if util.IsStringEmpty(string(newGroup.Config.Strategy.Type)) {
+		_ = render.Render(w, r, newErrorResponse("please provide a valid strategy type", http.StatusBadRequest))
+		return
+	}
```
danvixent, line 67: is this necessary since we're still going to compare to config.DefaultStrategyProvider?

server/group.go
Dotunj: I think this improves the experience of the API since we'll be returning the correct validation message.

danvixent: Yeah, true.

danvixent (approved):

PR #335, opened 2022-01-10
Configure Max response size to store

### LINKED ISSUE #328: Truncate endpoint response to a configurable limit.

danvixent:
Fixes #328

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -11,9 +11,12 @@ import (
 	"time"
 
 	"github.com/frain-dev/convoy/config/algo"
+	log "github.com/sirupsen/logrus"
 )
 
 var cfgSingleton atomic.Value
+
+const MaxResponseSize = 50 * 1024
 
 type DatabaseConfiguration struct {
 	Dsn string `json:"dsn"`
@@ -86,6 +89,7 @@ type Configuration struct {
 	Server            ServerConfiguration    `json:"server"`
 	Strategy          StrategyConfiguration  `json:"strategy"`
 	Signature         SignatureConfiguration `json:"signature"`
+	MaxResponseSize   uint64                 `json:"max_response_size"`
 	SMTP              SMTPConfiguration      `json:"smtp"`
 	Environment       string                 `json:"env"`
 	DisableEndpoint   bool                   `json:"disable_endpoint"`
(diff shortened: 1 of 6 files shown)


jirevwe (approved):

config/config.go
```diff
@@ -254,6 +258,16 @@ func LoadConfig(p string) error {
 		}
 	}
 
+	kb := c.MaxResponseSize * 1024 // to kilobyte
+	if kb == 0 {
+		c.MaxResponseSize = MaxResponseSize
+	} else if kb > MaxResponseSize {
+		log.Warnf("maximum response size of %dkb too large, using default value of 50kb", c.MaxResponseSize)
```
Subomi, line 265: @danvixent This error message seems off.

net/dispatcher.go
```diff
@@ -71,9 +71,12 @@ func (d *Dispatcher) SendRequest(endpoint, method string, jsonData json.RawMessa
 	}
 	updateDispatchHeaders(r, response)
 
-	body, err := ioutil.ReadAll(response.Body)
-	r.Body = body
-	if err != nil {
+	body := make([]byte, maxResponseSize)
+	n, err := io.ReadFull(response.Body, body)
+	r.Body = body[:n]
```
Subomi, line 76: @danvixent Any special reason why we're slicing here?

danvixent: we need to trim it down to the size that was actually written. if not the rest of body will contain 0s - because maxResponseSize won't always be reached

config/config.go
danvixent: hm, c.MaxResponseSize is passed in from config in kb terms. How do you suggest we make it better.

Subomi: I mean It should be `log.Warnf("maximum response size of %dkb is too large, using default value of %dkb", kb, c.MaxResponseSize)`

net/dispatcher.go
Subomi: If `maxResponseSize` is not reached, line 75 should error according to the [docs](https://pkg.go.dev/io#ReadFull). Did you notice a different behaviour in your test?

config/config.go
danvixent: Nah, kb is the resolved into bytes equivalent of c.MaxResponseSize, and MaxResponseSize(the const) is also in bytes.
We want to print it in kb

net/dispatcher.go
danvixent: Yes, it returns io.EOF if fewer bytes were read, we ignore EOF.

config/config.go
Subomi: Got it. You can change the `50kb` to use the const -- `MaxResponseSize` though.

net/dispatcher.go
danvixent: Noted, I ignore EOF which would happen if resp body is empty, will also ignore UnExpectedEOF that will occur if len(resp.Body) < len(body) but resp.Body will be drained.

config/config.go
danvixent: done

net/dispatcher.go
Subomi: @danvixent This doesn't solve the problem. `r.Body` will be empty if you only ignore `UnExpectedEOF`.

PR #339, opened 2022-01-11
feat: add helper for decoding json request payload

### LINKED ISSUE #333: Correct Error Message for Malformed JSON Payload
At the moment, a generic error message is sent back to the client whenever there's something wrong with the request payload. To improve the experience of the API, we can return back to the client the appropriate error message when any of the following occurs:

- Empty Request Body
- Malformed Request Body
- Incorrect JSON type for a request field

Dotunj:
Resolves #333


util/json.go
```diff
@@ -19,3 +25,35 @@ func JsonReMarshalString(s string) (string, error) {
...
+			if unmarshalTypeError.Field != "" {
+				return fmt.Errorf("body contains incorrect JSON type for field %q", unmarshalTypeError.Field)
+			}
+			return fmt.Errorf("body contains incorrect JSON type (at character %d)", unmarshalTypeError.Offset)
+
+		case errors.Is(err, io.EOF):
+			return errors.New("body must not be empty")
+
+		case errors.As(err, &invalidUnMarshalError):
+			panic(err)
```
jirevwe, line 51: Do we have a global panic recoverer? i thought it was bad practice to panic

jirevwe (approved):

util/json.go
danvixent: I think we do, but we shouldn't panic like this. @Dotunj 👀

Dotunj: I don't think it's necessarily a bad thing to panic at this stage. If we get the` invalidUnMarshalError` at runtime, it is most likely an error on the developer's part indicating that the wrong type was passed to `Decode()`

PR #341, opened 2022-01-12
feat: add library for handling request validation

### LINKED ISSUE #334: Improve Request Validation
Use a third party package to handle request validation. Wrap all validation errors together and return to client at once.

Dotunj:
Resolves #334 #290


util/validator.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func init() {
+	govalidator.TagMap["supported_hash"] = govalidator.Validator(func(hash string) bool {
+		if _, ok := algo.M[hash]; !ok {
+			return false
+		}
+
+		return true
+	})
```
danvixent, line 38: I think this can be somewhere in main.go

Dotunj: I thought about doing that too, but my reason for doing this is just incase we need to add more custom validation down the line, it's cleaner and clearer this way. Any downsides to doing it this way? @danvixent

config/config.go
```diff
@@ -67,18 +67,18 @@ type AuthConfiguration struct {
 }
 
 type StrategyConfiguration struct {
-	Type    StrategyProvider             `json:"type"`
-	Default DefaultStrategyConfiguration `json:"default"`
+	Type    StrategyProvider             `json:"type" valid:"required~please provide a valid strategy type,in(default)~unsupported strategy type"`
```
Subomi, line 70: @Dotunj This looks like we're coupling configuration validation with server validation. I don't think we want to do this.

server/group_test.go
```diff
@@ -121,7 +121,22 @@ func TestApplicationHandler_CreateGroup(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
-	bodyReader := strings.NewReader(`{"name": "ABC_DEF_TEST_UPDATE"}`)
+	bodyReader := strings.NewReader(`{
```
Subomi, line 124: @Dotunj There should be more tests for this guy. Like catch validation errors using a test table.

util/validator.go
jirevwe: I prefer it here, let's not litter main with so much logic.

jirevwe (approved):

util/validator.go
Subomi: This is probably fine.

Subomi (changes requested):
FYI @Dotunj

config/config.go
Dotunj: @Subomi Are you referring to just this particular field?

PR #342, opened 2022-01-12
Add In-Memory Queue Support

ogbanugot:
This pr adds support for in-memory queue configuration.


worker/producer.go
```diff
@@ -15,8 +17,25 @@ type Producer struct {
...
+			consumer = queue.Consumer()
+		}
+	}
+
+	if cfg.Queue.Type == config.InMemoryQueueProvider {
+		if queue, ok := eventQueue.(*convoyMemqueue.MemQueue); ok {
+			consumer = queue.Consumer()
+			err := consumer.Stop()
+			if err != nil {
+				log.Fatal(err)
```
danvixent, line 35: why are we stopping the consumer?

danvixent: hm, we stop it here. only to start it again somewhere else?

ogbanugot: we start it as usual in cmd/server.go.

queue/storage.go
```diff
@@ -0,0 +1,46 @@
...
+	"sync"
+
+	"github.com/hashicorp/golang-lru/simplelru"
+)
+
+type Storage interface {
+	Exists(ctx context.Context, key string) bool
+}
+
+var _ Storage = (*localStorage)(nil)
```
jirevwe, line 14: What does this line do? Pardon/Indulge my lack of Go knowledge

jirevwe (approved):

queue/storage.go
ogbanugot: > What does this line do? Pardon/Indulge my lack of Go knowledge

The `_` in front of the variable tells the compiler to effectively discard the RHS value, but to type-check it and evaluate it if it has any side effects, but the anonymous `Storage` variable per se doesn't take any process space. It's basically a way to ensure that localstorage implements type Storage at compile time.

queue/queue.go
```diff
@@ -6,15 +6,28 @@ import (
...
+	"github.com/go-redis/redis/v8"
+	"github.com/vmihailenco/taskq/v3"
 )
 
 type Queuer interface {
 	io.Closer
 	Write(context.Context, convoy.TaskName, *datastore.EventDelivery, time.Duration) error
 }
 
+type QueueClient interface {
```
Subomi, line 20: @ogbanugot I don't think we want to do this. What we want to do is:
- Create the client.
- Pass it to the `NewQueue` function for the respective queue.
 
This is implementing standard DI.

queue/queue.go
```diff
@@ -6,15 +6,28 @@ import (
...
+	NewClient(config.Configuration) (*StorageClient, taskq.Factory, error)
+	NewQueue(StorageClient, taskq.Factory, string) Queuer
+}
+
 type Job struct {
 	Err error  `json:"err"`
 	ID  string `json:"id"`
 }
+
+type StorageClient struct {
```
Subomi, line 30: @ogbanugot What we will rather do here is to create our own `QueueOptions`. This will contain a subset of fields from `taskq.QueueOptions` and will be passed into `NewQueue`.

queue/queue.go
```diff
@@ -6,15 +6,28 @@ import (
 	"time"
```
Subomi, line 6: @ogbanugot This file should define something like this

```
func NewQueue(opts QueueOptions) Queuer {
  if opts.Type == "in-memory" {
	return memqueue.NewQueue(opts)
  }

  if opts.Type == "redis" {
	return redisqueue.NewQueue(opts)
  }

  return errors.New("invalid type")
}
```

This is what will be called in `main.go` as `queue.NewQueue(opts)`

queue/storage.go
```diff
@@ -0,0 +1,46 @@
+package queue
```
Subomi, line 1: @ogbanugot Do we need to copy this guy? I think we can keep the interface but get `taskq.NewLocalStorage()`

Please correct me if I'm wrong.

worker/producer.go
```diff
@@ -15,8 +17,25 @@ type Producer struct {
 	quit          chan chan error
 }
 
-func NewProducer(queue *convoyRedis.RedisQueue) *Producer {
-	consumer := queue.Consumer()
+func NewProducer(cfg config.Configuration, eventQueue queue.Queuer) *Producer {
+	var consumer taskq.QueueConsumer
+	var queue queue.Queuer
+
+	if cfg.Queue.Type == config.RedisQueueProvider {
```
Subomi, line 24: @ogbanugot This should be wrapped into the `Queuer` interface.

Subomi (changes requested):
FYI @ogbanugot

cmd/main.go
```diff
@@ -170,6 +170,18 @@ func main() {
...
+func getQueueClient(cfg config.Configuration) queue.QueueClient {
+	var queueClient queue.QueueClient
+	if cfg.Queue.Type == config.RedisQueueProvider {
+		queueClient = convoyRedis.NewQueueClient()
+	}
+	if cfg.Queue.Type == config.InMemoryQueueProvider {
+		queueClient = convoyMemqueue.NewQueueClient()
+	}
+
+	return queueClient
```
oneEyedSunday, line 182: Should we make allowance for an error here, or does the caller have to deduce that on their own?

queue/queue.go
ogbanugot: Can QueueOptions be defined here though? It will cause an import cycle.

cmd/main.go
ogbanugot: I'm currently refactoring.

worker/producer.go
ogbanugot: This is not very clear.

queue/queue.go
ogbanugot: Makes sense, this takes care of the check for which `NewQueue()` to call, what about the check for which `NewClient()` to call, should we set a condition on the `cfg.Queue.Type` in `cmd/main.go`. In other words where does the toggle between queue types happen in `cmd/main.go`?

queue/queue.go
ogbanugot: Also, I think `memqueue` and `redisqueue` can both return an instance of our own `QueueOptions` so that we can toggle between the two in `cmd/main.go` with ease.

queue/memqueue/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			appID := uuid.NewString()
+			eventID := uuid.NewString()
+			eventDeliveryID := uuid.NewString()
+
+			evenDelivery := &datastore.EventDelivery{
```
Subomi, line 35: Little typo here. `evenDelivery`

queue/memqueue/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+			if err != nil {
+				t.Fatalf("Failed to get config")
+
+			}
+
+			var qFn taskq.Factory
+			var lS queue.Storage
+			var opts queue.QueueOptions
+
+			if cfg.Queue.Type == config.InMemoryQueueProvider {
```
Subomi, line 61: This is in-memory test file, is there any reason why we're doing this check here again?

queue/redis/client_test.go
```diff
@@ -0,0 +1,105 @@
...
+					Name:    uuid.NewString(),
+					Type:    "redis",
+					Redis:   rC,
+					Factory: qFn,
+				}
+			}
+
+			eventQueue := NewQueue(opts)
+			switch tc.testType {
+			case "writer":
```
Subomi, line 81: We should have two separate test functions for the `Write` & `Consumer` and use a test table to check for multiple scenarios for each method.

queue/memqueue/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+					Name:    uuid.NewString(),
+					Type:    "in-memory",
+					Storage: lS,
+					Factory: qFn,
+				}
+			}
+
+			eventQueue := NewQueue(opts)
+			switch tc.testType {
+			case "writer":
```
Subomi, line 76: Same thing here.

Subomi (changes requested):
Nice. To fix the broken CI we should

- Add a Redis instance to `.github/workflows/go.yml` 
- Add a new step: `Run Redis Integration tests` 
- Specify integration tags on the Redis queue test file.

ogbanugot:
> Nice. To fix the broken CI we should
> 
> * Add a Redis instance to `.github/workflows/go.yml`
> * Add a new step: `Run Redis Integration tests`
> * Specify integration tags on the Redis queue test file.

 I noticed it was running everything twice, so I merged the two steps as one integration step

queue/queue.go
```diff
@@ -7,14 +7,29 @@ import (
...
 	"github.com/frain-dev/convoy"
 	"github.com/frain-dev/convoy/datastore"
+	"github.com/go-redis/redis/v8"
+	"github.com/vmihailenco/taskq/v3"
 )
 
 type Queuer interface {
 	io.Closer
 	Write(context.Context, convoy.TaskName, *datastore.EventDelivery, time.Duration) error
+	Consumer() taskq.QueueConsumer
```
danvixent, line 17: @Subomi @ogbanugot should we be returning external objects in our own interfaces?

queue/redis/client_test.go
```diff
@@ -0,0 +1,124 @@
...
+func TestRedisQueue(t *testing.T) {
+	for scenario, fn := range map[string]func(t *testing.T){
+		"redis queue write":    testWritetoQueue,
+		"redis start consumer": testStartConsumer,
+	} {
+		t.Run(scenario, func(t *testing.T) {
+			fn(t)
+		})
+	}
+}
```
danvixent, line 29: i don't know if this kind of tabling will allow us efficiently test various scenarios.

Please correct me if I'm wrong.

ogbanugot: Can you clarify your concerns?

danvixent: oh i mean, i mean table test scenarios for testWritetoQueue for example, like [here](https://github.com/frain-dev/convoy/blob/main/datastore/bolt/application_test.go)

ogbanugot: Yeah, i was doing something like this initially `https://github.com/frain-dev/convoy/pull/342/commits/6047d604aad3f20837764809d438aa7979423dd5` but @Subomi requested a different approach where the tests are separated into unique functions and then called in the test table.

Subomi: @ogbanugot This is not how we write table tests in the codebase. You can look through tests in the `server` package to understand how to correct it.

ogbanugot: Okay, will make changes.

Subomi (approved):

lzap:
Hey, I am trying to use the in-memory queue but it appears to require redis:

```
		factory := memqueue.NewFactory()
		queue.JobQueue = factory.RegisterQueue(&taskq.QueueOptions{
			Name:        config.QueueName(),
			Storage:     taskq.NewLocalStorage(),
			WorkerLimit: 1,
		})
```

I am running into:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x38 pc=0x1037afb70]

goroutine 79 [running]:
github.com/bsm/redislock.(*Client).obtain(0x140002b5260, {0x10475ca20, 0x140000ac960}, {0x140001a35f0, 0x24}, {0x14000b31c68?, 0x1056c8300?}, 0x64?)
        /Volumes/SSD/Go/pkg/mod/github.com/bsm/redislock@v0.7.2/redislock.go:101 +0x50
github.com/bsm/redislock.(*Client).Obtain(0x140002b5260, {0x10475c9e8, 0x140001a4000}, {0x140001a35f0, 0x24}, 0x1037cb70c?, 0x0)
        /Volumes/SSD/Go/pkg/mod/github.com/bsm/redislock@v0.7.2/redislock.go:73 +0x1b8
github.com/bsm/redislock.Obtain({0x10475c9e8, 0x140001a4000}, {0x0?, 0x0}, {0x140001a35f0, 0x24}, 0x14000084688?, 0x102388310?)
        /Volumes/SSD/Go/pkg/mod/github.com/bsm/redislock@v0.7.2/redislock.go:129 +0x98
github.com/vmihailenco/taskq/v3.(*Consumer).lockWorker(0x1400013dc70, {0x10475c9e8, 0x140001a4000}, 0x0, 0x2101?)
        /Volumes/SSD/Go/pkg/mod/github.com/vmihailenco/taskq/v3@v3.2.9/consumer.go:720 +0x170
github.com/vmihailenco/taskq/v3.(*Consumer).worker(0x1400013dc70, {0x10475c9e8?, 0x140001a4000}, 0x0)
        /Volumes/SSD/Go/pkg/mod/github.com/vmihailenco/taskq/v3@v3.2.9/consumer.go:468 +0xd8
github.com/vmihailenco/taskq/v3.(*Consumer).addWorker.func1()
        /Volumes/SSD/Go/pkg/mod/github.com/vmihailenco/taskq/v3@v3.2.9/consumer.go:276 +0x64
created by github.com/vmihailenco/taskq/v3.(*Consumer).addWorker
        /Volumes/SSD/Go/pkg/mod/github.com/vmihailenco/taskq/v3@v3.2.9/consumer.go:274 +0x1c0
```

Is it expected behavior?

PR #345, opened 2022-01-13
Configure Max response size to store

danvixent:
replicates #335 for main

### DIFF
--- a/net/dispatcher_test.go
+++ b/net/dispatcher_test.go
@@ -146,7 +146,7 @@ func TestDispatcher_SendRequest(t *testing.T) {
 				defer deferFn()
 			}
 
-			got, err := d.SendRequest(tt.args.endpoint, tt.args.method, tt.args.jsonData, tt.args.signatureHeader, tt.args.hmac)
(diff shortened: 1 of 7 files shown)


jirevwe (approved):

net/dispatcher_test.go
```diff
@@ -146,7 +146,7 @@ func TestDispatcher_SendRequest(t *testing.T) {
 				defer deferFn()
 			}
 
-			got, err := d.SendRequest(tt.args.endpoint, tt.args.method, tt.args.jsonData, tt.args.signatureHeader, tt.args.hmac)
+			got, err := d.SendRequest(tt.args.endpoint, tt.args.method, tt.args.jsonData, tt.args.signatureHeader, tt.args.hmac, config.MaxResponseSize)
```
Subomi, line 149: @danvixent We should have a test case to catch reading more bytes than configured.

danvixent: done

jirevwe:
Cc @Subomi is this ready?

PR #346, opened 2022-01-13
Refactored root directory

Subomi:
This pr removes data models and repository definitions in the root directory that constantly creates import cycles. 

The problem is solving the import cycle by bringing more things outside to the root was going to litter the root directory unnecessarily. So I moved everything into the datastore package. So the models, repository & their implementations can live together.


ogbanugot (commented):
Looks good. What if you want a file to store some types or const for use in cmd/*.go? That aren't specific to the existing packages?

Subomi:
> Looks good. What if you want a file to store some types or const for use in cmd/*.go? That aren't specific to the existing packages?

Yeah, I think we can create `types.go` in `cmd` then.

jirevwe (approved):

ogbanugot (approved):

ogbanugot:
> > Looks good. What if you want a file to store some types or const for use in cmd/*.go? That aren't specific to the existing packages?
> 
> Yeah, I think we can create `types.go` in `cmd` then.

Perfect.

danvixent (approved):

PR #347, opened 2022-01-13
Add applications bolt implementation

jirevwe:


datastore/bolt/application.go
```diff
@@ -0,0 +1,116 @@
...
+	return nil
+}
+
+func (a *appRepo) CountGroupApplications(tx context.Context, groupID string) (int64, error) {
+	return 0, nil
+}
+
+func (a *appRepo) createUpdateApplication(app *convoy.Application) error {
+	return a.db.Update(func(tx *bbolt.Tx) error {
+		b := tx.Bucket([]byte(a.bucketName))
```
danvixent, line 101: I figured we can still use constant for this thing

```go
const eventDeliveryBucketName = "eventdeliveries"
const eventBucketName = "events"

```

datastore/bolt/group.go
```diff
@@ -0,0 +1,114 @@
...
+	var group *convoy.Group
+	err := g.db.View(func(tx *bbolt.Tx) error {
+		b := tx.Bucket([]byte(g.bucketName))
+
+		grp := b.Get([]byte(gid))
+		if grp == nil {
+			return fmt.Errorf("group with id (%s) does not exist", gid)
+		}
+
+		var _grp *convoy.Group
```
danvixent, line 78: you ca reference the group you declared previously.

```go
err := json.Unmarshal(buf, group)
if err != nil {
   return err
}

```

datastore/bolt/application.go
jirevwe: Nah, just add it to the struct since we'd also need it to create the bucket

danvixent: e no go work 😭 , apparently you can't have multiple root buckets, but you can have multiple sub buckets in the root buckets. check my pr for reference.

cmd/main.go
```diff
@@ -87,6 +88,11 @@ func main() {
 				return err
 			}
 
+			bolt, err := bolt.New(cfg)
```
Subomi, line 91: @jirevwe we should define `datastore.NewDB(cfg)`. It will do the if/else and return `datastore.DatabaseClient`

Subomi (dismissed):

cmd/main.go
danvixent: i see a import cycle here, bolt -> datastore.Application, datastore -> bolt.New

jirevwe: Hmm, let me check

datastore/bolt/application.go
```diff
@@ -0,0 +1,287 @@
...
+		if pErr != nil {
+			return pErr
+		}
+
+		return nil
+	})
+}
+
+func (a *appRepo) DeleteGroupApps(ctx context.Context, gid string) error {
+	return a.db.View(func(tx *bbolt.Tx) error {
```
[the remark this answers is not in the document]
jirevwe: Good catch.

cmd/main.go
jirevwe: Yep, there's an import cycle. What do you think @Subomi?

Dotunj (approved):

datastore/bolt/group.go
```diff
@@ -1,114 +1,134 @@
...
-		mErr := json.Unmarshal(grp, &_grp)
-		if mErr != nil {
-			return mErr
+		var temp *datastore.Group
+		err := json.Unmarshal(grp, &temp)
+		if err != nil {
+			return err
 		}
-		group = _grp
+		group = temp
```
danvixent, line 82: you can just reference the `var group *datastore` here 
```go
err := json.Unmarshal(grp, &group)
if err != nil {
    return err
}
```

datastore/bolt/group.go
```diff
@@ -1,114 +1,134 @@
...
 	err := g.db.View(func(tx *bbolt.Tx) error {
-		id := name + ":" + gid
+		b := tx.Bucket([]byte(g.bucketName))
 
-		grp := tx.Bucket([]byte(bucketName)).Get([]byte(id))
+		grp := b.Get([]byte(gid))
 		if grp == nil {
-			return fmt.Errorf("group with id (%s) does not exist", gid)
+			return datastore.ErrGroupNotFound
 		}
```
danvixent, line 75: we can call this `buf` instead as that name denotes the byte slice property better

datastore/bolt/application.go
```diff
@@ -3,70 +3,319 @@ package bolt
...
+			data.TotalPage = int64(math.Ceil(float64(b.Stats().KeyN) / float64(perPage)))
+			data.Total = int64(b.Stats().KeyN)
+		} else {
+			total, err := a.CountGroupApplications(ctx, gid)
+			if err != nil {
+				return nil
+			}
+
+			data.TotalPage = int64(math.Ceil(float64(total) / float64(perPage)))
+			data.Total = int64(total)
```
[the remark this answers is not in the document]
jirevwe: This is not correct though. I added a test case to validate it and it's not correct

danvixent (approved):

datastore/bolt/application.go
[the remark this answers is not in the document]
danvixent: nicee.

PR #366, opened 2022-01-18
Add Partial Index with Unique Constraint for Group Name

### LINKED ISSUE #363: Ensure partial index on group names
Currently, group names are unique fields in the group collection. However, with soft-deletes, we cannot create new groups with the same name. :( 

We should set partial indexes on group names with a condition to filter active groups.

Dotunj:
Resolves #363


danvixent (approved):

danvixent:
@Dotunj is it possible to have a test case for this?

Dotunj:
@danvixent Yeah, added a test case

datastore/mongo/group_test.go
```diff
@@ -52,6 +54,53 @@ func Test_CreateGroup(t *testing.T) {
 	require.Equal(t, org.UID, newOrg.UID)
 }
 
+func Test_CannotCreateGroupWithExistingName(t *testing.T) {
```
Subomi, line 57: @Dotunj This should be a test table under `Test_CreateGroup`

Subomi (changes requested):

Subomi (approved):

PR #387, opened 2022-01-24
Register task handlers for groups created over the API

ogbanugot:
This schedules a routine that checks through the `groupRepo` and registered a task handler for newly created groups.


danvixent (approved):

oneEyedSunday:
Why so many commits though?

Subomi (approved):

PR #390, opened 2022-01-24
feat: add api key bolt implementation

Dotunj:


datastore/bolt/api_key.go
```diff
@@ -2,54 +2,249 @@ package bolt
...
 import (
 	"context"
+	"encoding/json"
+	"math"
 
 	"github.com/frain-dev/convoy/datastore"
 	"go.etcd.io/bbolt"
 )
 
+const apiKeyBucketName = "apiKeys"
```
jirevwe, line 12: Is there a reason why you didn't use the struct field in the apiKeyRepo?

datastore/bolt/api_key_test.go
```diff
@@ -0,0 +1,303 @@
...
+					Name:  "Group 1",
+					Count: 10,
+				},
+			},
+			expected: Expected{
+				ApiKeyCount:    3,
+				paginationData: datastore.PaginationData{Total: 10, TotalPage: 4, Page: 1, PerPage: 3, Prev: 0, Next: 2},
+			},
+		},
+	}
```
jirevwe, line 235: I don't think test tables are useful if you're only planning to add one test case

datastore/bolt/api_key_test.go
```diff
@@ -0,0 +1,303 @@
...
+	db, closeFn := getDB(t)
+	defer closeFn()
+
+	apiKeyRepo := NewApiRoleRepo(db)
+
+	maskID, salt, encodedKey, err := generateAPIKey()
+
+	if err != nil {
+		t.Fatal(err)
+	}
```
jirevwe, line 62: Why not use `require.NoError()`
```suggestion
		require.NoError(t, err)
```

datastore/bolt/api_key.go
```diff
@@ -2,54 +2,249 @@ package bolt
...
-	return nil
+func (a *apiKeyRepo) FindAPIKeyByHash(ctx context.Context, hash string) (*datastore.APIKey, error) {
+	var apiKeys []datastore.APIKey = make([]datastore.APIKey, 0, 1)
+	var apiKey datastore.APIKey
+
+	err := a.db.View(func(tx *bbolt.Tx) error {
+		b := tx.Bucket([]byte(a.bucketName))
+		c := b.Cursor()
+
+		for k, v := c.First(); k != nil; k, v = c.Next() {
```
jirevwe, line 137: This can be replaced with `b.ForEach()`, I think we try to use the provided abstractions unless we really need to do some custom stuff with the foo-loop

datastore/bolt/api_key_test.go
jirevwe, line 235: I don't think test tables are useful if you're only planning to add one test case

jirevwe (changes requested):
Probably add one more test case for the paged function

datastore/bolt/api_key.go
```diff
@@ -2,54 +2,249 @@ package bolt
...
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+}
+
+func (a *apiKeyRepo) UpdateAPIKey(ctx context.Context, apiKey *datastore.APIKey) error {
+	return a.createUpdateApplication(apiKey)
```
danvixent, line 55: `createUpdateAPIKey`

datastore/bolt/api_key.go
Dotunj: struct field is still being used. Just assigned a constant for the bucket name

datastore/bolt/api_key.go
danvixent: I think seeking works here, for each will execute for every single pair in the db even after we find what we want.

datastore/bolt/api_key.go
```diff
@@ -2,54 +2,249 @@ package bolt
...
+
+	if pageable.PerPage < 1 {
+		perPage = 10
+	}
+
+	if page < 1 {
+		prevPage = 1
+	} else {
+		prevPage = page - 1
+	}
```
danvixent, line 183: The if..else isn't necessary, just `prevPage = page - 1` is enough

datastore/bolt/api_key.go
```diff
@@ -2,54 +2,249 @@ package bolt
...
+		b := tx.Bucket([]byte(a.bucketName))
+		c := b.Cursor()
+		i := 1
+
+		for k, v := c.First(); k != nil; k, v = c.Next() {
+			if v == nil {
+				continue
+			}
+
+			if i > perPage*prevPage && i <= perPage*page {
```
danvixent, line 195: i tihnk storing the bounds makes sense here, instead of having computed on every iteration,
```go
lowerBound := perPage*prevPage
upperBound := perPage*page
```

datastore/bolt/api_key.go
```diff
@@ -2,54 +2,249 @@ package bolt
...
+				apiKeys = append(apiKeys, apiKey)
+			}
+			i++
+
+			if i == (perPage*page)+perPage {
+				break
+			}
+		}
+
+		total, err := a.countAPIKeys()
```
danvixent, line 211: @jirevwe b.Stats().KeyN should suffice here right?

datastore/bolt/api_key_test.go
Dotunj: I don't think that assertion suits this test case, since it's something a bit different.

jirevwe: Hmm, Ok

datastore/bolt/api_key.go
```diff
@@ -2,54 +2,222 @@ package bolt
...
+		return nil
+	})
+
+	return err
 }
 
-func (e *apiKeyRepo) FindAPIKeyByHash(ctx context.Context, hash string) (*datastore.APIKey, error) {
-	return nil, nil
+func (a *apiKeyRepo) FindAPIKeyByHash(ctx context.Context, hash string) (*datastore.APIKey, error) {
+	var apiKeys []datastore.APIKey = make([]datastore.APIKey, 0, 1)
```
jirevwe, line 130: Is the use of an array here really necessary? On line 160 we're only accessing the first element anyway

jirevwe (approved):

PR #400, opened 2022-01-26
Feat: use contains instead of get / error handling

obbap1:
This PR addresses two things:
1. In `localStorage` we use `Get` and then check if the key exists in the LRU cache. We can use `Contains` to just peek if the key exists. 
2. return errors/nil if there is no error, no need for extra checks.

### DIFF
--- a/queue/storage.go
+++ b/queue/storage.go
@@ -13,15 +13,18 @@ type Storage interface {
 
 var _ Storage = (*localStorage)(nil)
 
+const defaultStorageSize = 128000
+
 // LOCAL
 
 type localStorage struct {
 	mu    sync.Mutex
 	cache *simplelru.LRU
+	size  int
 }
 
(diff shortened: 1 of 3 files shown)


queue/storage.go
```diff
@@ -30,14 +33,13 @@ func (s *localStorage) Exists(_ context.Context, key string) bool {
...
 		if err != nil {
 			panic(err)
 		}
 	}
 
-	_, ok := s.cache.Get(key)
-	if ok {
+	if ok := s.cache.Contains(key); ok {
 		return true
 	}
```
danvixent, line 44: this can be:
```go
return s.cache.Contains(key)
```
right?

obbap1: yes

ogbanugot (approved):

danvixent (approved):

PR #422, opened 2022-02-01
Add CLI for Queue Observability

ogbanugot:
- View info about the queue through a CLI tool.

### DIFF
--- a/cmd/queue.go
+++ b/cmd/queue.go
@@ -69,17 +69,12 @@ func getZSETLength(a *app) *cobra.Command {
 		Short: "get ZSET Length",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			q := a.eventQueue.(*redisqueue.RedisQueue)
-			redisStreamClient := q.Inner
 			ctx := context.Background()
 			ticker := time.NewTicker(time.Duration(timeInterval) * time.Millisecond)
-			zset := "taskq:" + "{" + q.Name + "}:zset"
 			for {
 				select {
 				case <-ticker.C:
-					bodies, err := redisStreamClient.ZRangeByScore(ctx, zset, &redis.ZRangeBy{
-						Min: "-inf",
-						Max: "+inf",
-					}).Result()
+					bodies, err := q.ZRangebyScore(ctx, "-inf", "+inf")
 					if err != nil {
 						log.Printf("Error ZSET Length: %v", err)
 					}
@@ -103,13 +98,11 @@ func getStreamInfo(a *app) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
 			q := a.eventQueue.(*redisqueue.RedisQueue)
-			redisStreamClient := q.Inner
 			ticker := time.NewTicker(time.Duration(timeInterval) * time.Millisecond)
-			stream := "taskq:" + "{" + q.Name + "}:stream"
 			for {
 				select {
 				case <-ticker.C:
-					r, err := redisStreamClient.XInfoStream(ctx, stream).Result()
+					r, err := q.XInfoStream(ctx).Result()
 					if err != nil {
 						log.Printf("XInfoStream err: %v", err)
 					}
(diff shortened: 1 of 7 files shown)


queue/memqueue/client.go
```diff
@@ -16,7 +16,7 @@ import (
 type MemQueue struct {
 	Name      string
 	queue     *memqueue.Queue
-	inner     taskq.Factory
+	Inner     taskq.Factory
```
oneEyedSunday, line 19: Why would you want to make `inner` public? By convention its supposed to be hidden right?

cmd/queue.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+//get length of ZSET, delayed msgs
+func getZSETLength(a *app) *cobra.Command {
+	var timeInterval int
+	cmd := &cobra.Command{
+		Use:   "zsetlength",
+		Short: "get ZSET Length",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			q := a.eventQueue.(*redisqueue.RedisQueue)
```
oneEyedSunday, line 71: Question, does this technically break if we use another queue client?
It feels like it would.

Do we also just want to encapsulate and expose methods for the range in the interface of the queue itself? as against the implementation assertions here?

Or are we certain that we'd only ever use a redis queue (and then, need observability against the redis queue implementation only)?

cmd/queue.go
```diff
@@ -0,0 +1,382 @@
...
+			redisStreamClient := q.Inner
+			ctx := context.Background()
+			ticker := time.NewTicker(time.Duration(timeInterval) * time.Millisecond)
+			zset := "taskq:" + "{" + q.Name + "}:zset"
+			for {
+				select {
+				case <-ticker.C:
+					bodies, err := redisStreamClient.ZRangeByScore(ctx, zset, &redis.ZRangeBy{
```
oneEyedSunday, line 79: This should probably be encapsulated in a method on the queue, where this is the redis queue implementation

cmd/queue.go
```diff
@@ -0,0 +1,382 @@
...
+// Get general stream info
+func getStreamInfo(a *app) *cobra.Command {
+	var timeInterval int
+	cmd := &cobra.Command{
+		Use:   "streaminfo",
+		Short: "get stream info",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+			q := a.eventQueue.(*redisqueue.RedisQueue)
+			redisStreamClient := q.Inner
```
oneEyedSunday, line 106: same for this too, this way we don't have to expose `queue.inner`

queue/memqueue/client.go
ogbanugot: It's the cleanest way to access the redis client, we could change the name from `inner`

cmd/queue.go
ogbanugot: yes it would break stating that that `a.eventQueue` is not of type `redis`. A command that has `zset` in it is clearly for redis queue only. Also in terms of encapsulating the commands here, taskq already encapsulates these commands behind the RedisStreamClient which is an interface of type `redis.Client ` hence my need to access the redis client through "inner".

cmd/queue.go
```diff
@@ -0,0 +1,404 @@
...
+					if value == id {
+						inPending = true
+					}
+				case "msg":
+					if msg.ID == id {
+						inPending = true
+					}
+				}
+
+			}
```
danvixent, line 382: is it possible to have this logic in a separate func?

cmd/queue.go
```diff
@@ -0,0 +1,404 @@
...
+					value := string(msg.ArgsBin[argsSlice:])
+					if value == id {
+						inZSET = true
+					}
+				case "msg":
+					if msg.ID == id {
+						inZSET = true
+					}
+				}
+			}
```
danvixent, line 312: is it possible to have this logic in a separate func?

cmd/queue.go
```diff
@@ -0,0 +1,404 @@
...
+					return err
+				}
+				switch idType {
+				case "eventdev":
+					value := string(msg.ArgsBin[argsSlice:])
+					if value == id {
+						onQueue = true
+					}
+				case "msg":
+					if msg.ID == id {
```
danvixent, line 249: is it possible to have this logic in a separate func?

danvixent (approved):

cmd/queue.go
ogbanugot: the check in the Switch statement?

ogbanugot: Based on our conversation earlier, I've done some clean up to help make those components more reusable. Still working on a delete function.

cmd/queue.go
```diff
@@ -0,0 +1,309 @@
...
+					log.Printf("Queue Length: %+v\n", length)
+				case <-ctx.Done():
+					return nil
+				}
+			}
+		},
+	}
+	cmd.Flags().IntVar(&timeInterval, "interval", 2000, "Log time interval")
+	return cmd
+}
```
jirevwe, line 59: I think it'll be helpful to get the queue length in terms of "capacity". This function might return 200, but I won't understand what the number means, does 200 mean it's full? Or does it mean it's half full. I'm only suggesting this if it's possible.

jirevwe (approved):

PR #424, opened 2022-02-01
add apps badger implementation

Dotunj:


datastore/bolt/application.go
```diff
@@ -132,240 +84,131 @@ func (a *appRepo) LoadApplicationsPagedByGroupId(ctx context.Context, gid string
...
-		var app *datastore.Application
-		err := json.Unmarshal(appBytes, &app)
-		if err != nil {
-			return err
-		}
 
-		for _, v := range app.Endpoints {
-			if v.UID == endpointID {
-				endpoint = &v
+	err := a.db.ForEach(badgerhold.Where("UID").Eq(appID), func(application *datastore.Application) error {
```
danvixent, line 116: A better option would be to fetch the app using a.db.Get(app.UID) and then iterate it's endpoints looking for the one with that endpoint id

datastore/bolt/application.go
```diff
@@ -132,240 +84,131 @@ func (a *appRepo) LoadApplicationsPagedByGroupId(ctx context.Context, gid string
...
+
 	return endpoint, err
 }
 
 func (a *appRepo) DeleteApplication(ctx context.Context, app *datastore.Application) error {
-	return a.db.Update(func(tx *bbolt.Tx) error {
-		b := tx.Bucket([]byte(a.bucketName))
-		return b.Delete([]byte(app.UID))
-	})
+	return a.db.DeleteMatching(&datastore.Application{}, badgerhold.Where("UID").Eq(app.UID))
```
danvixent, line 134: Here i think we can delete with `a.store.Delete(app.UID, app)`

datastore/bolt/application.go
```diff
@@ -132,240 +84,131 @@ func (a *appRepo) LoadApplicationsPagedByGroupId(ctx context.Context, gid string
...
+				return false, fmt.Errorf("Field not a string, it's a %T!", field)
 			}
 
-			return nil
-		})
-	})
-
-	return i, err
-}
+			return strings.Contains(strings.ToLower(field.(string)), strings.ToLower(f.title)), nil
```
danvixent, line 195: qFunc("Title").Contains() will suffice here

danvixent: seen. thanks

PR #431, opened 2022-02-04
add support for cache layer

Dotunj:
This PR introduces a cache layer for convoy.  Supported types are `redis` and the default type is `in-memory`
For frequently accessed data or for expensive DB queries such as the graph on the dashboard, we can store and retrieve that data from a cache to significantly improve response time. However, the downside to this, is that from time to time, the data being displayed on the graph might be stale.


jirevwe (approved):

cache/memory/client.go
```diff
@@ -0,0 +1,50 @@
...
+	if errors.Is(err, cache.ErrCacheMiss) {
+		return nil, nil
+	}
+
+	if err != nil {
+		return err, nil
+	}
+
+	return nil, data
+}
```
danvixent, line 46: Hm, i think this signature can be better. Like so:
```go
func (m *MemoryCache) Get(ctx context.Context, key string, data interface{}) error {}
// OR
func (m *MemoryCache) Get(ctx context.Context, key string) (error, interface{}) {}
```

cache/redis/client.go
```diff
@@ -0,0 +1,59 @@
...
+	if errors.Is(err, cache.ErrCacheMiss) {
+		return nil, nil
+	}
+
+	if err != nil {
+		return err, nil
+	}
+
+	return nil, data
+}
```
danvixent, line 55: Hm, i think this signature can be better. Like so:
```go
func (m *MemoryCache) Get(ctx context.Context, key string, data interface{}) error {}
// OR
func (m *MemoryCache) Get(ctx context.Context, key string) (error, interface{}) {}
```

ogbanugot (approved):

danvixent (approved):

PR #439, opened 2022-02-05
check for nil app endpoints

danvixent:


jirevwe (approved):

Dotunj (approved):

datastore/badger/application.go
```diff
@@ -66,12 +66,17 @@ func (a *appRepo) LoadApplicationsPaged(ctx context.Context, gid, q string, page
 		return nil, datastore.PaginationData{}, err
 	}
 
-	for i, ap := range apps {
+	for i := range apps {
```
oneEyedSunday, line 69: If the ap.Endpoints is nil, can we skip the a.db.Count query since its expected to be zero?

danvixent: I think that would be an incorrect assumption, what if the endpoints were deleted after events have been pushed through the app?

oneEyedSunday: I guess my question is, is it correct to assume that if there are no application endpoints, there are no application events?

danvixent: No i don't think so. Deletions of either may have occured.

oneEyedSunday:
Ok

PR #494, opened 2022-03-03
feat: add taskq incremental autotune

ogbanugot:


type.go
```diff
@@ -53,3 +53,12 @@ const (
...
 )
+
+const (
+	// Maximum number of goroutines fetching messages.
+	MaxNumFetcher = 100
+	// Number of messages reserved by a fetcher in the queue in one request.
+	ReservationSize = 1000
+	//Size of the internal buffer
+	BufferSize = 100000
+)
```
jirevwe, line 64: We should be able to set these as env var later. @ogbanugot

jirevwe (approved):

Dotunj (approved):

danvixent (approved):

PR #511, opened 2022-03-08
Prefer environment variables over convoy config file

jirevwe:
This PR allows convoy be configured using environment variables which would take preference over values that have been set in the `convoy.json` file.


config/realm.go
```diff
@@ -1,14 +1,40 @@
...
+	*b = config
+	return err
+}
+
+type APIKeyAuthConfig []APIKeyAuth
+
 type APIKeyAuth struct {
 	APIKey string    `json:"api_key"`
 	Role   auth.Role `json:"role"`
 }
```
danvixent, line 31: This takes away the ability to set multiple right?

danvixent (approved):

config/realm.go
jirevwe: Nah, we can still set mutiple like this 
```sh
export CONVOY_BASIC_AUTH_CONFIG="[{\"username\": \"some-admin\",\"password\": \"some-password\",\"role\": {\"type\": \"super_user\",\"groups\": []}}]"
export CONVOY_API_KEY_CONFIG="[{\"api_key\":\"ABC1234\",\"role\":{\"type\":\"admin\",\"groups\":[\"group-uid-1\",\"group-uid-2\"],\"apps\":[\"apps-uid-1\",\"apps-uid-2\"]}}]"
```

ogbanugot (approved):

PR #528, opened 2022-03-16
Add timestamp to header

ogbanugot:
- This pr adds timestamp to headers to allow checks for replay attacks.

### DIFF
--- a/util/crypto.go
+++ b/util/crypto.go
@@ -8,10 +8,8 @@ import (
 	"crypto/sha512"
 	"encoding/hex"
 	"errors"
-	"fmt"
 	"hash"
 	"strings"
-	"time"
 
 	"github.com/dchest/uniuri"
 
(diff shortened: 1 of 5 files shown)


jirevwe (approved):

util/crypto.go
```diff
@@ -8,21 +8,23 @@ import (
...
-const (
-	prefix    string = "CO"
-	seperator string = "."
+var (
+	prefix    []string = []string{"CO", "t=", "data="}
+	seperator []string = []string{".", ","}
 )
 
-func ComputeJSONHmac(hash, data, secret string, order bool) (string, error) {
+func ComputeJSONHmac(hash, data, secret string, order bool, timestamp bool) (string, error) {
```
jirevwe, line 27: I think this should be the default, no need for the boolean check

danvixent (approved):

PR #545, opened 2022-03-23
Fix: Add delay to task handler errors

ogbanugot:
- pass `context.background()` when writing to queue
- add delay to all task handler errors

### DIFF
--- a/services/event_service.go
+++ b/services/event_service.go
@@ -134,7 +134,7 @@ func (e *EventService) CreateAppEvent(ctx context.Context, newMessage *models.Ev
 		taskName := convoy.EventProcessor.SetPrefix(g.Name)
 
 		if eventDelivery.Status != datastore.DiscardedEventStatus {
-			err = e.eventQueue.Write(ctx, taskName, eventDelivery, 1*time.Second)
+			err = e.eventQueue.Write(context.Background(), taskName, eventDelivery, 1*time.Second)
 			if err != nil {
 				log.Errorf("Error occurred sending new event to the queue %s", err)
 			}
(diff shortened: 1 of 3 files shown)


services/event_service.go
```diff
@@ -134,7 +134,7 @@ func (e *EventService) CreateAppEvent(ctx context.Context, newMessage *models.Ev
 		taskName := convoy.EventProcessor.SetPrefix(g.Name)
 
 		if eventDelivery.Status != datastore.DiscardedEventStatus {
-			err = e.eventQueue.Write(ctx, taskName, eventDelivery, 1*time.Second)
+			err = e.eventQueue.Write(context.Background(), taskName, eventDelivery, 1*time.Second)
```
jirevwe, line 137: Oh, did I miss this one in my other PR?

worker/task/process_event_delivery.go
```diff
@@ -44,9 +44,12 @@ func ProcessEventDelivery(appRepo datastore.ApplicationRepository, eventDelivery
 		// Load message from DB and switch state to prevent concurrent processing.
 		m, err := eventDeliveryRepo.FindEventDeliveryByID(context.Background(), Id)
 
+		var delayDuration time.Duration
```
jirevwe, line 47: Don't you think we can define and assign this once instead of reassigning it multiple times?

jirevwe (changes requested):

services/event_service.go
ogbanugot: oh seen.

jirevwe: Please don't undo it, I might have missed them

worker/task/process_event_delivery.go
ogbanugot: the mock tests where failing when I did it that way but let me dig deeper and see.

jirevwe: Hmm, I think is cos you need to EXPECT the call to the repo function

ogbanugot: yeah exactly, fixed it it now thanks.

worker/task/process_event_delivery.go
```diff
@@ -46,8 +46,9 @@ func ProcessEventDelivery(appRepo datastore.ApplicationRepository, eventDelivery
 
 		if err != nil {
 			log.WithError(err).Errorf("Failed to load event - %s", Id)
-			return nil
+			return &EndpointError{Err: err}
```
jirevwe, line 49: @ogbanugot pass the delay here too

jirevwe (changes requested):

PR #547, opened 2022-03-23
add group service layer

danvixent:


services/group_service.go
```diff
@@ -0,0 +1,153 @@
...
+	err = gs.appRepo.DeleteGroupApps(ctx, id)
+	if err != nil {
+		log.WithError(err).Error("failed to delete group apps")
+		return NewServiceError(http.StatusInternalServerError, errors.New("failed to delete group apps"))
+	}
+
+	err = gs.eventRepo.DeleteGroupEvents(ctx, id)
+	if err != nil {
+		log.WithError(err).Error("failed to delete group events")
+		return NewServiceError(http.StatusInternalServerError, errors.New("failed to delete group events"))
```
jirevwe, line 149: Can we use this opportunity to change all the occurrences of this status code (for the files in this PR alone)  
```suggestion
		return NewServiceError(http.StatusBadRequest, errors.New("failed to delete group events"))
```

jirevwe (changes requested):

services/group_service.go
danvixent: Point taken. I will implement this.

ogbanugot (approved):

Dotunj (approved):

PR #614, opened 2022-04-21
updated core dashboard loaders

Oluwadaminiola:


web/ui/dashboard-component/projects/convoy-dashboard/src/lib/loader-component/table-loader.component.ts
```diff
@@ -0,0 +1,104 @@
...
+					<td>
+						<div>
+							<div class="skeleton-loader data"></div>
+						</div>
+					</td>
+					<td *ngFor="let head of tableHead.slice(1)">
+						<div class="skeleton-loader data"></div>
+					</td>
+				</tr>
+			</tbody>
```
horlah, line 95: I dropped a comment on similar implementation of this on SaaS, please reference it.

horlah (changes requested):

web/ui/dashboard-component/projects/convoy-dashboard/src/lib/loader-component/table-loader.component.ts
Oluwadaminiola: i already did a workaround here for this number of data appearing on each row

horlah (approved):

PR #628, opened 2022-04-27
Set span in ProcessEventDelivery

ogbanugot:
- Sets a span in ProcessEventDelivery using newrelic tracer, so that we can have an estimate of the average time it takes to process an event delivery.

### DIFF
--- a/worker/task/process_event_delivery.go
+++ b/worker/task/process_event_delivery.go
@@ -50,8 +50,8 @@ func ProcessEventDelivery(appRepo datastore.ApplicationRepository, eventDelivery
 		var txn *newrelic.Transaction
 		if tracer != nil {
 			txn = tracer.StartTransaction("ProcessEventDelivery")
+			defer txn.End()
 		}
-		defer txn.End()
 
 		Id := job.ID
 
--- a/worker/task/process_event_delivery.go
+++ b/worker/task/process_event_delivery.go
@@ -16,6 +16,7 @@ import (
 	"github.com/frain-dev/convoy/net"
 	"github.com/frain-dev/convoy/queue"
 	"github.com/frain-dev/convoy/retrystrategies"
+	"github.com/frain-dev/convoy/tracer"
 	"github.com/frain-dev/convoy/util"
 	"github.com/google/uuid"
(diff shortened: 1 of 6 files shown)


jirevwe (approved):

worker/task/process_event_delivery.go
```diff
@@ -43,8 +45,14 @@ type SignatureValues struct {
...
 }
 
-func ProcessEventDelivery(appRepo datastore.ApplicationRepository, eventDeliveryRepo datastore.EventDeliveryRepository, groupRepo datastore.GroupRepository, rateLimiter limiter.RateLimiter) func(*queue.Job) error {
+func ProcessEventDelivery(appRepo datastore.ApplicationRepository, eventDeliveryRepo datastore.EventDeliveryRepository, groupRepo datastore.GroupRepository, rateLimiter limiter.RateLimiter, tracer tracer.Tracer) func(*queue.Job) error {
 	return func(job *queue.Job) error {
+		var txn *newrelic.Transaction
+		if tracer != nil {
+			txn = tracer.StartTransaction("ProcessEventDelivery")
+		}
+		defer txn.End()
```
danvixent, line 54: Should this guy also be in the if condition?

jirevwe:
@ogbanugot do you think we need to add a span for `ProcessEventCreation()` too?

danvixent:
> @ogbanugot do you think we need to add a span for ProcessEventCreation() too?

I personally think we should @jirevwe

worker/task/process_event_delivery.go
ogbanugot: No, if we put it in the condition won't the transaction be scoped to the if condition?

danvixent: Yes it would, because if the condition is false, txn will be `nil` and calling txn.End will panic

jirevwe:
> > @ogbanugot do you think we need to add a span for ProcessEventCreation() too?
> 
> I personally think we should @jirevwe

I agree, though it's out of scope of this PR. Plus I think we have all the data we need in the event delivery record, hmm.

danvixent (approved):

PR #635, opened 2022-04-29
prevent duplicate app names

### LINKED ISSUE #607: Multiple applications with the same name created
It's possible to create multiple applications with the same name.

image

danvixent:
fixes #607


datastore/mongo/mongo.go
```diff
@@ -113,6 +113,7 @@ func (c *Client) ensureMongoIndices() {
 	c.ensureIndex(EventCollection, "event_type", false, nil)
 	c.ensureIndex(EventCollection, "app_metadata.uid", false, nil)
 	c.ensureIndex(AppCollections, "group_id", false, nil)
+	c.ensureIndex(AppCollections, "title", true, nil)
```
Dotunj, line 116: I think this index should also be scoped to the `group_id` along with the `document_status`. That way multiple apps under different groups can still have the same name, and then we can still reuse the same name for an app that has been deleted.

services/app_service.go
```diff
@@ -36,6 +36,19 @@ func (a *AppService) CreateApp(ctx context.Context, newApp *models.Application,
 		return nil, NewServiceError(http.StatusBadRequest, err)
 	}
 
+	apps, _, err := a.LoadApplicationsPaged(ctx, "", newApp.AppName, datastore.Pageable{
```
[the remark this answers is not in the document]
danvixent: i think this is a better option 👍

services/app_service.go
```diff
@@ -49,7 +62,7 @@ func (a *AppService) CreateApp(ctx context.Context, newApp *models.Application,
 		DocumentStatus:  datastore.ActiveDocumentStatus,
 	}
 
-	err := a.appRepo.CreateApplication(ctx, app)
+	err = a.appRepo.CreateApplication(ctx, app)
```
Dotunj, line 65: Since we already have an index in place, `err` will not be nil for non-unique names. Is it possible to type assert the value of `err` at this stage and return the correct error message back to the client?

danvixent: In hindsight, since we are now calling `IsAppTitleUnique` i will be removing the unique index

datastore/mongo/mongo.go
danvixent: > In hindsight, since we are now calling IsAppTitleUnique i will be removing the unique index

Subomi: Please let's use the index. Indexes are the last line of defence for data issues. I prefer Indexes over software checks. The actual advantage software checks give us over indexes is human-readable messages, but we can transform index errors into human-readable messages with a little more work.

danvixent: That would make the behaviors of our database layers uneven - badger doesn't have unique indexes.
But yes we can extract the error details using `mongo.IsDuplicateKeyError` and [this](https://github.com/mongodb/mongo-go-driver/blob/master/mongo/errors.go#L276:6), the only problem here is the details would be awkward to return to the frontend.

We can leave the software check & keep the index for mongo, from the service layer perspective the behaviors are still even.

datastore/repository.go
```diff
@@ -49,6 +49,7 @@ type GroupRepository interface {
 type ApplicationRepository interface {
 	CreateApplication(context.Context, *Application) error
 	LoadApplicationsPaged(context.Context, string, string, Pageable) ([]Application, PaginationData, error)
+	IsAppTitleUnique(context.Context, string, string) (bool, error)
```
Subomi, line 52: @danvixent Hmm. Why isn't this just `IsAppTitleUnique(context.Context, string, string) error`. This simplifies the API and the test where we are checking for `error` and the `bool` value.

Subomi (changes requested):
@danvixent FYI

datastore/mongo/mongo.go
Subomi: @danvixent This is fine. Let's keep both for mongo & only software checks for badger. Realise also that we don't currently recommend badger for prod but people are running mongo in prod so we need to be more careful.

datastore/repository.go
danvixent: I agree that the definition is awkward, and i personally don't like it, but in this case, i think it is necessary, we are trying to check if an app name is unique => `true` or `false`, `error` reports any error we face along the way.

datastore/mongo/mongo.go
danvixent: That works 👍

datastore/repository.go
danvixent: @Subomi

datastore/badger/application.go
```diff
@@ -23,10 +23,28 @@ func NewApplicationRepo(db *badgerhold.Store) datastore.ApplicationRepository {
 }
 
 func (a *appRepo) CreateApplication(ctx context.Context, app *datastore.Application) error {
+	unique, err := a.isAppTitleUnique(ctx, app)
```
Subomi, line 26: @danvixent Change `func isAppTitleUnique(ctx, app) (bool, error)` to `func assertUniqueAppTitle(ctx, app) error` let it return errors or `datastore.ErrDuplicateAppName` and remove the duplicate checks here.

Subomi (approved):

jirevwe (changes requested):

jirevwe (approved):

PR #638, opened 2022-04-30
Integration tests

Subomi:


jirevwe:
I thought we wanted to use immune to do the integration tests @Subomi 🤔

Subomi:
@jirevwe No, we use immune for acceptance tests.

Issue #639, opened 2022-05-01
Endpoint `Secrets` are not being updated


CreamyMilk:
Apparently when I update an endpoint's `Secret` Convoy doesn't persist the update.

Demo of the issue 👉 : https://github.com/CreamyMilk/convoy-stuff/blob/main/main.go

oneEyedSunday:
We might probably also want to check if this affects other resources (since it seems we're having to self select which properties get updated on some entities)

Subomi:
This issue was a feature, not a bug. :(  The reasoning was secrets are sensitive things, and we wouldn't want them to be updated "easily". But we didn't add a new API for it (e.g. `PUT /applications/{appID}/endpoints/{endpointID}/secret`), which leaves the user with the last choice of deleting the old app and creating a new one with the new secret -- Not the best solution. 

Thanks for raising the issue & pr.

PR #640, opened 2022-05-01
fix/Update endpoints secret

CreamyMilk:
This PR ensures that an  endpoints `Secret` are persisted when updated 

Fix for : #639


Subomi (approved):
LGTM.

jirevwe (approved):

PR #652, opened 2022-05-10
Search implementation

jirevwe:
This PR introduces [Typesense](https://typesense.org/) as a search backend. Documents are indexed using a change stream which can be seen [here](https://github.com/frain-dev/indexer/blob/master/index.js#L60-L70).

Configuration options were added to allow users configure a search backend. MongoDB or Badger (Disk) will be used to search if one isn't supplied.

Events can be searched by accessing the get many events endpoint `GET /v1/events?query=xxx`


ogbanugot (approved):

searcher/typesense/client.go
```diff
@@ -0,0 +1,97 @@
...
+
+	filterByBuilder.WriteString(fmt.Sprintf("app_metadata.group_id:=%s", f.Group.UID))
+
+	hasAppFilter := !util.IsStringEmpty(f.AppID)
+	if hasAppFilter {
+		filterByBuilder.WriteString(fmt.Sprintf(" && app_metadata.uid:=%s", f.Group.UID))
+	}
+
+	filterByBuilder.WriteString(fmt.Sprintf(" && created_at:[%d..%d]", f.SearchParams.CreatedAtStart*1000, f.SearchParams.CreatedAtEnd*1000))
+
```
danvixent, line 47: Hm, i'm confused, why multiply by 1000?

server/event.go
```diff
@@ -252,19 +253,47 @@ func (a *applicationHandler) ForceResendEventDeliveries(w http.ResponseWriter, r
...
+		if err != nil {
+			_ = render.Render(w, r, newErrorResponse(err.Error(), http.StatusBadRequest))
+			return
+		}
+		_ = render.Render(w, r, newServerResponse("App events fetched successfully",
+			pagedResponse{Content: &m, Pagination: &paginationData}, http.StatusOK))
+
+		return
+	}
+
```
danvixent, line 296: I thought we only redirect to typesense when `query` isn't empty?

searcher/typesense/client.go
jirevwe: Because the search middleware converts epoch in seconds but we need them in milliseconds

server/event.go
jirevwe: Nope, there's no point hitting the db if we already have a search backend indexing documents we can query from

danvixent: okay

searcher/typesense/client.go
danvixent: i see.

danvixent (approved):

PR #658, opened 2022-05-11
build projects modal

horlah:


web/ui/dashboard-component/projects/convoy-dashboard/src/lib/convoy-dashboard.component.scss
```diff
@@ -63,3 +63,23 @@
 		height: fit-content;
 	}
 }
+
```
Oluwadaminiola, line 66: this can be in the ui library and some of these classes (e.g max-width and padding) are already existing classes in the UI library

horlah: Yea, I don't want us to put styles that belong to just one project in the library. Things in the library should be styles used in two or more projects.

I've added the width and padding styles with the UI library class name though. @Oluwadaminiola

Oluwadaminiola (approved):

PR #659, opened 2022-05-11
New sidebar design for app/event delivery details

Oluwadaminiola:


horlah:
@Oluwadaminiola Well done with this. Please review these:
App and event delivery details are two separate things, so they should be in two different components. It might be helpful to understand that these components will end up being a page when we move the dashboard component back into part of the core angular project.
- You should display the dashboard by default, then you can use a query parameter of a variable name to set/know to should the details page
- overview-component is confusing please; please use app-details/event-delivery-details

@Oluwadaminiola

Oluwadaminiola:
> 

okay noted

horlah (approved):

PR #663, opened 2022-05-12
adds support for storing events for disabled apps

### LINKED ISSUE #650: Disable apps should persist events.
Currently, if you send events to a disabled application, the API responds with a 400 Bad Request and drops the event on the floor. Applications can be disabled for two reasons: 
1. The system automatically disabled it because of consistent delivery issues.
2. The Operator turned the application off. 

Both reasons could be to debug a transient issue. In this scenario, Convoy should not drop events on the floor. Our priority is reliability and ease of use without requiring the provider to push these events again from their systems.

Dotunj:
Resolves #650

### DIFF
--- a/services/event_service.go
+++ b/services/event_service.go
@@ -76,10 +76,6 @@ func (e *EventService) CreateAppEvent(ctx context.Context, newMessage *models.Ev
 		return nil, NewServiceError(http.StatusBadRequest, errors.New("app has no configured endpoints"))
 	}
 
-	if app.IsDisabled {
-		return nil, NewServiceError(http.StatusBadRequest, errors.New("app is disabled, no events were sent"))
-	}
-
 	event := &datastore.Event{
 		UID:       uuid.New().String(),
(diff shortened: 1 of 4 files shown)


services/event_service.go
```diff
@@ -76,10 +76,6 @@ func (e *EventService) CreateAppEvent(ctx context.Context, newMessage *models.Ev
 		return nil, NewServiceError(http.StatusBadRequest, errors.New("app has no configured endpoints"))
 	}
 
-	if app.IsDisabled {
-		return nil, NewServiceError(http.StatusBadRequest, errors.New("app is disabled, no events were sent"))
-	}
-
```
danvixent, line 82: I we should still check this somewhere. If it app is disabled we should save the event and create event deliveries for it, but those event deliveries should not enter the queue.

cc @Subomi

Dotunj: That's the current implementation in this PR.

jirevwe (approved):

danvixent (approved):

PR #672, opened 2022-05-18
add ReplayAppEvent endpoint

### LINKED ISSUE #660: Manually replay events
Currently, the retry button only works for Event Deliveries. If for some reason an event's event deliveries do not get created, there is no way to trigger them.

danvixent:
fixes #660


server/route.go
```diff
@@ -125,6 +125,7 @@ func buildRoutes(app *applicationHandler) http.Handler {
 				eventRouter.Route("/{eventID}", func(eventSubRouter chi.Router) {
 					eventSubRouter.Use(requireEvent(app.eventRepo))
 					eventSubRouter.Get("/", app.GetAppEvent)
+					eventSubRouter.Put("/", app.ReplayAppEvent)
```
Subomi, line 128: Let's make this `PUT /replay`

services/event_service_test.go
```diff
@@ -486,6 +486,74 @@ func TestEventService_GetAppEvent(t *testing.T) {
...
+	}
+	tests := []struct {
+		name        string
+		args        args
+		dbFn        func(es *EventService)
+		wantErr     bool
+		wantErrCode int
+		wantErrMsg  string
+	}{
+		{
```
Subomi, line 504: This test case doesn't specify `wantErr` as `false`.

Subomi (changes requested):
@danvixent FYI

services/event_service_test.go
danvixent: fixed

server/route.go
danvixent: done

server/middleware.go
```diff
@@ -456,6 +456,7 @@ func requireGroup(groupRepo datastore.GroupRepository, cache cache.Cache) func(n
 					}
 				}
 			} else {
+				// TODO(all): maybe we should only use default-group if require_auth is false?
```
jirevwe, line 459: I think we should not even have the `require_auth` config option

danvixent: @Subomi see this

Subomi: Yes, I agree. Let's eliminate it. @danvixent

Subomi: Obviously in another PR. @danvixent

danvixent: okay

Subomi (approved):

PR #688, opened 2022-05-27
Ingestion API

Subomi:
This PR introduces the API for ingestion events from external services.


server/ingest.go
```diff
@@ -0,0 +1,99 @@
...
+	log "github.com/sirupsen/logrus"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func (a *applicationHandler) IngestEvent(w http.ResponseWriter, r *http.Request) {
+	// 1. Retrieve mask ID
+	maskID := r.URL.Query().Get("maskID")
+
+	// 2. Retrieve source using mask ID.
+	source, err := a.sourceRepo.FindSourceByMaskID(r.Context(), maskID)
```
Dotunj, line 24: After retrieving the source, do we need to verify that the `source_type` is actually a `http` type to be able to proceed further?

server/ingest.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		_ = render.Render(w, r, newErrorResponse(err.Error(), http.StatusBadRequest))
+		return
+	}
+
+	// 3. Select verifier based of source config.
+	var v verifier.Verifier
+	verifierConfig := source.Verifier
```
Dotunj, line 32: do we need to assert that the `verifier` field on the source is not `nil` before proceeding with the verification?

server/ingest.go
jirevwe: I think we do, because if it was changed verification will fail. On the flip side when updating sources the config should only be changed and not swapped out.

jirevwe (approved):

PR #691, opened 2022-05-30
Setup cron system

Subomi:


danvixent (approved):

Dotunj (approved):

cmd/scheduler.go
```diff
@@ -24,40 +21,16 @@ func addSchedulerCommand(a *app) *cobra.Command {
...
-				case <-ctx.Done():
-					ticker.Stop()
-					return
-				}
-			}
+			// Register tasks.
+			// s.AddTask("retry events", 30, func() {
+			// 	  task.RetryEventDeliveries(nil, "", a.eventDeliveryRepo, a.groupRepo, a.eventQueue)
+			// })
+
```
jirevwe, line 31: Why are these commented out though?

jirevwe (approved):

cmd/scheduler.go
Subomi: Because we don't need it again? I left it as a template for the next jobs we'll be adding.

jirevwe: Oh ok

PR #694, opened 2022-05-30
Update process event delivery tests

jirevwe:
This PR fixes the all the tests that were broken when we introduced the use of subscriptions to map between event sources and apps/endpoints.


datastore/mongo/subscription.go
```diff
@@ -152,3 +152,20 @@ func (s *subscriptionRepo) FindSubscriptionBySourceIDs(ctx context.Context, grou
 
 	return subscription, nil
 }
+
+func (s *subscriptionRepo) UpdateSubscriptionsStatus(ctx context.Context, groupId string, subscriptionId string, status datastore.EndpointStatus) error {
```
danvixent, line 156: I think this should be `UpdateSubscriptionStatus`

server/testdb/seed.go
```diff
@@ -276,6 +278,49 @@ func SeedSource(db datastore.DatabaseClient, g *datastore.Group, uid string) (*d
...
+	filterConfig *datastore.FilterConfiguration,
+) (*datastore.Subscription, error) {
+	if util.IsStringEmpty(uid) {
+		uid = uuid.New().String()
+	}
+
+	subscription := &datastore.Subscription{
+		UID:        uid,
+		GroupID:    g.UID,
+		Name:       "",
```
danvixent, line 299: Shouldn’t `name` also be an param of this func?

services/event_service.go
```diff
@@ -231,25 +232,21 @@ func (e *EventService) RetryEventDelivery(ctx context.Context, eventDelivery *da
...
+	if sub.Status == datastore.PendingEndpointStatus {
+		return errors.New("subscription is being re-activated")
+	}
 
-	// 	err = e.appRepo.UpdateApplicationEndpointsStatus(context.Background(), eventDelivery.AppMetadata.UID, pendingEndpoints, datastore.PendingEndpointStatus)
-	// 	if err != nil {
-	// 		return errors.New("failed to update endpoint status")
-	// 	}
-	// }
+	if sub.Status == datastore.InactiveEndpointStatus {
```
danvixent, line 244: Same thing here

services/event_service.go
```diff
@@ -259,15 +256,14 @@ func (e *EventService) forceResendEventDelivery(ctx context.Context, eventDelive
...
-	// }
+	sub, err := e.subRepo.FindSubscriptionByID(ctx, g.UID, eventDelivery.SubscriptionID)
+	if err != nil {
+		return ErrSubscriptionNotFound
+	}
 
-	// if endpoint.Status != datastore.ActiveEndpointStatus {
-	// 	return errors.New("force resend to an inactive or pending endpoint is not allowed")
-	// }
+	if sub.Status != datastore.ActiveEndpointStatus {
```
danvixent, line 264: Here too

services/event_service_test.go
```diff
@@ -1483,16 +1500,18 @@ func TestEventService_RetryEventDelivery(t *testing.T) {
 				g: &datastore.Group{UID: "abc"},
 			},
 			wantErr:    true,
-			wantErrMsg: "endpoint is being re-activated",
+			wantErrMsg: "subscription is being re-activated",
 		},
 		{
 			name: "should_retry_event_delivery_with_inactive_endpoint",
```
danvixent, line 1506: Should this change to `should_retry_event_delivery_with_inactive_subscription`? as with other test cases here.

datastore/mongo/subscription.go
jirevwe: Ops, updated it. Thanks

server/testdb/seed.go
jirevwe: Well the names are pretty useless in the context of the tests, so I didn't bother adding it

danvixent: Hm, okay cool

datastore/mongo/subscription.go
danvixent: Nice

PR #697, opened 2022-06-01
Tie Users into organisations

danvixent:


services/organisation_invite_service.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, NewServiceError(http.StatusBadRequest, errors.New("failed to create organisation member invite"))
+	}
+
+	user := &datastore.User{
+		UID:       uuid.NewString(),
+		FirstName: newUser.FirstName,
+		LastName:  newUser.LastName,
+		Email:     email,
+		Password:  string(p.Hash),
+		//Role:          newUser.Role, // TODO(all): this role field shouldn't be in user.
```
jirevwe, line 138: Isn't there a "role" on the OrgMember schema already?

server/models/models.go
```diff
@@ -94,6 +103,15 @@ type Token struct {
 	AccessToken  string `json:"access_token" valid:"required~please provide an access token"`
 	RefreshToken string `json:"refresh_token" valid:"required~please provide a refresh token"`
 }
+
+type User struct {
```
Dotunj, line 107: We should add validation for some of the required fields for this struct

server/organisation_invite.go
```diff
@@ -0,0 +1,84 @@
...
+// @Param email query string true "email"
+// @Param accepted query string true "email"
+// @Param user body models.User false "User Details"
+// @Success 200 {object} serverResponse{data=Stub}
+// @Failure 400,401,500 {object} serverResponse{data=Stub}
+// @Security ApiKeyAuth
+// @Router /process_organisation_member_invite [post]
+func (a *applicationHandler) ProcessOrganisationMemberInvite(w http.ResponseWriter, r *http.Request) {
+	token := r.URL.Query().Get("token")
+	email := r.URL.Query().Get("email")
```
Dotunj, line 61: Why do we need the `email` here? I'm guessing the `token` on its own should be sufficient.

server/route.go
```diff
@@ -76,6 +76,8 @@ func buildRoutes(app *applicationHandler) http.Handler {
 			r.Use(jsonResponse)
 			r.Use(requireAuth())
 
+			r.Post("/process_organisation_member_invite", app.ProcessOrganisationMemberInvite)
```
Dotunj, line 79: Not sure this route follows the `REST` pattern.

services/organisation_member_service.go
```diff
@@ -0,0 +1,94 @@
...
+type OrganisationMemberService struct {
+	orgMemberRepo datastore.OrganisationMemberRepository
+}
+
+func NewOrganisationMemberService(orgMemberRepo datastore.OrganisationMemberRepository) *OrganisationMemberService {
+	return &OrganisationMemberService{orgMemberRepo: orgMemberRepo}
+}
+
+func (om *OrganisationMemberService) CreateOrganisationMember(ctx context.Context, org *datastore.Organisation, user *datastore.User, role *auth.Role) (*datastore.OrganisationMember, error) {
+	err := role.Validate("organisation member")
```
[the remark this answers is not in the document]
danvixent: Yeah, i agree. This can be done in a new PR.

server/route.go
jirevwe: True, we'd need to create an invites resource for that then

server/organisation_invite.go
danvixent: You're right. I have removed it.

server/route.go
Dotunj: `/organisations/members/accept-invite` as an example and it can be under the `/ui` route.

Dotunj (approved):

PR #700, opened 2022-06-02
Add missing comma to convoy.json.example

djperrefort:
I was running the convoy example and ran into the following error:

```
Error: invalid character '"' after object key:value pair
```

It looks like there was a comma missing in the example configuration file.


ozombo (commented):
👏

PR #716, opened 2022-06-07
auth integration

Oluwadaminiola:
In this pr, the authentication endpoints were integrated and the request service was updated, to test, login with email: `superuser@default.com` and password: `default`. You also need to update your `convoy.json` file with this `"jwt": {
            "enabled": true
        },` under your `auth` object


horlah (approved):
Please review again and space the code before merging

PR #723, opened 2022-06-08
add user settings endpoints

Dotunj:


services/user_service.go
```diff
@@ -148,4 +148,57 @@ func (u *UserService) token() (*jwt.Jwt, error) {
...
+func (u *UserService) UpdateUser(ctx context.Context, data *models.UpdateUser, user *datastore.User) (*datastore.User, error) {
+	if err := util.Validate(data); err != nil {
+		return nil, NewServiceError(http.StatusBadRequest, err)
+	}
+
+	user.FirstName = data.FirstName
+	user.LastName = data.LastName
+	user.Email = data.Email
+
+	err := u.userRepo.UpdateUser(ctx, user)
```
Subomi, line 162: @Dotunj If I mistakenly pass in the `Password` here, will it update the password?

Subomi (changes requested):
FYI

services/user_service.go
Dotunj: Pass in `Password` from the request payload? No, the password won't be updated.

danvixent (approved):

Subomi (approved):

PR #725, opened 2022-06-08
add document_status filter to all find queries

### LINKED ISSUE #693: Scope all MongoDB queries with document_status
Filter all mongo db queries with the document_status field to prevent deleted documents from turning up.

danvixent:
fixes #693


Subomi (approved):

datastore/mongo/mongo_test.go
```diff
@@ -14,7 +13,7 @@ import (
 )
 
 func getDSN() string {
-	return os.Getenv("TEST_MONGO_DSN")
+	return "mongodb://localhost:27017/convoyxx"
```
jirevwe, line 16: This should not be hardcoded

jirevwe (changes requested):

jirevwe (approved):

PR #742, opened 2022-06-10
Feat: Forgot Password

ogbanugot:


internal/email/templates/reset.password.html
```diff
@@ -0,0 +1,123 @@
...
+                            </p>
+
+                            <p class="issue-text">
+                                For any enquiry or complaint, you can reply to this email.
+                            </p>
+                        </div>
+
+                        <div class="center footer">
+                            <p>© <a href="https://getconvoy.io">Convoy</a></p>
+                            <p>A Cloud native Webhook Service</p>
```
jirevwe, line 116:
```suggestion
                            <p>An OSS Webhook Service</p>
```

services/user_service.go
```diff
@@ -202,3 +212,93 @@ func (u *UserService) UpdatePassword(ctx context.Context, data *models.UpdatePas
 
 	return user, nil
 }
+
+func (u *UserService) GeneratePasswordResetToken(ctx context.Context, baseURL string, data *models.ForgotPassword) error {
+	var resetToken string
```
jirevwe, line 217: why is reset token defined here?

jirevwe (changes requested):

services/user_service.go
ogbanugot: oh I'm using a variable because I pass it into the `sendPasswordResetEmail` function as well.

jirevwe: Yeah, but it's initialized on line 230 so there's not need to define it here

ogbanugot: okay I see your point.

Subomi (approved):

jirevwe (approved):

danvixent (approved):

PR #745, opened 2022-06-11
Chore: remove example scheduler task

ogbanugot:


cmd/scheduler.go
```diff
@@ -32,16 +30,16 @@ func addSchedulerCommand(a *app) *cobra.Command {
 			}
 			ctx := context.Background()
 
-			taskName := convoy.TaskName("example-schedule")
+			// taskName := convoy.TaskName("example-schedule")
```
jirevwe, line 33: Why not just delete these lines??

jirevwe (changes requested):

cmd/scheduler.go
ogbanugot: true

jirevwe (approved):

PR #778, opened 2022-06-20
validate app, endpoint & source when creating subscription

danvixent:


services/subscription_service.go
```diff
@@ -39,6 +39,22 @@ func (s *SubcriptionService) CreateSubscription(ctx context.Context, groupID str
 		return nil, NewServiceError(http.StatusBadRequest, err)
 	}
 
+	app, err := s.appRepo.FindApplicationByID(ctx, newSubscription.AppID)
```
Dotunj, line 42: We should add a similar check for the source ID too if it's present in the request.

Dotunj (approved):

services/subscription_service.go
Subomi: I mean, why not here as well? @danvixent

danvixent: Nothing really. I'll do it here

jirevwe (approved):

ogbanugot (approved):

PR #809, opened 2022-06-28
Ui Fixes

Oluwadaminiola:
In this pr, the analytics modal was updated and bug fixes were made for:
- creating a new organization for a new user
- updating a GitHub ingester type source
- fetching apps on create subscriptions component

### DIFF
--- a/web/ui/dashboard/src/app/private/private.component.html
+++ b/web/ui/dashboard/src/app/private/private.component.html
@@ -23,9 +23,9 @@
 				<span class="font__weight-400 color__primary">Go to docs</span>
 			</a>
 
-			<div class="dropdown" *ngIf="organisations?.length">
+			<div class="dropdown">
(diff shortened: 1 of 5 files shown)


jirevwe (approved):

web/ui/dashboard/src/app/private/private.component.html
```diff
@@ -23,9 +23,9 @@
 				<span class="font__weight-400 color__primary">Go to docs</span>
 			</a>
 
-			<div class="dropdown" *ngIf="organisations?.length">
```
horlah, line 26: The issue here is that this dropdown becomes available when the user doesn't have an organisation, it'll throw an error. @Oluwadaminiola

Oluwadaminiola: I moved the conditional so that the dropdown would show that there are no current organizations available and would make the `add organization` button available to the user even when they are not on the dashboard

PR #831, opened 2022-07-05
Created select component with tailwind

Oluwadaminiola:


web/ui/dashboard/tailwind.config.js
```diff
@@ -57,7 +57,8 @@ module.exports = {
 				80: '#31323D',
 				60: '#5F5F68',
 				40: '#737A91',
-				20: '#E8E8E9'
+				20: '#E8E8E9',
+				10: '#EDEDF5'
```
horlah, line 61: Where is this color from?

web/ui/dashboard/src/app/components/select/select.component.ts
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+  registerOnChange() {}
+
+	registerOnTouched() {}
+
+	writeValue() {}
+
+	setDisabledState() {}
+}
```
horlah, line 46: Review indentation on this file

web/ui/dashboard/src/app/components/select/select.component.html
```diff
@@ -0,0 +1,44 @@
...
+			hover:bg-primary-500
+			hover:border-grey-20
+			focus:border-primary-100
+			focus:bg-white-100
+			outline-none
+			rounded-4px
+			bg-[#F7F9FC]
+			py-12px
+			px-16px
+		"
```
horlah, line 35: Please test the different states of the styles here against the design system, hover, disabled and error state

horlah (changes requested):

web/ui/dashboard/tailwind.config.js
Oluwadaminiola: this is the color used for almost all borders and also used for the required tags

horlah (approved):

PR #844, opened 2022-07-07
created page component with tailwind

Oluwadaminiola:


web/ui/dashboard/src/app/components/page/page.component.ts
```diff
@@ -0,0 +1,23 @@
...
+	@Input('size') size: 'small' | 'normal' = 'normal';
+	@Input('class') class!: string;
+	sizes = { small: 'max-w-[848px]', normal: 'max-w-[1374px]' };
+
+	constructor() {}
+
+	ngOnInit(): void {}
+
+  get classes(): string{
+    return `${this.sizes[this.size]} ${this.class} ${this.size === 'small' ? 'bg-white-100 border border-grey-10 rounded-8px mt-10' : 'px-8 pb-8 pt-16'}`
```
horlah, line 21: Why do you have this condition here, why not set the classes here to the small value in the sizes object above?

horlah (changes requested):

horlah (approved):

PR #855, opened 2022-07-10
Updated private layout

Oluwadaminiola:


web/ui/dashboard/src/app/private/private.component.scss
```diff
@@ -1,15 +1,5 @@
-nav {
-	margin-left: 130px;
-
-	@media (max-width: 1150px) {
-		margin-left: 20px;
+nav{
+	.active{
+		@apply bg-primary-500 text-primary-100 font-medium;
```
horlah, line 3: Can you try see if there is a way to have this in the HTML?

horlah (changes requested):

horlah (approved):

PR #861, opened 2022-07-13
Updated sources with tailwind

Oluwadaminiola:
In this pr,
- the source page was updated with tailwind
- the create source component was updated with tailwind and prebuilt components
- the select and input components were updated with the tooltip components specs
- the tooltip component was updated with its accepted specs
- the radio component was updated to contain tooltips
- tooltip size spec was updated across board

### DIFF
--- a/web/ui/dashboard/src/app/private/components/create-source/create-source.component.ts
+++ b/web/ui/dashboard/src/app/private/components/create-source/create-source.component.ts
@@ -69,6 +69,8 @@ export class CreateSourceComponent implements OnInit {
 
 	async saveSource() {
 		const verifier = this.sourceForm.get('verifier.type')?.value === 'github' ? 'hmac' : this.sourceForm.get('verifier.type')?.value;
+		console.log('fvdf')
+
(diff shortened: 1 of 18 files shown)


web/ui/dashboard/src/app/private/components/create-source/create-source.component.ts
```diff
@@ -69,6 +69,8 @@ export class CreateSourceComponent implements OnInit {
 
 	async saveSource() {
 		const verifier = this.sourceForm.get('verifier.type')?.value === 'github' ? 'hmac' : this.sourceForm.get('verifier.type')?.value;
+		console.log('fvdf')
```
horlah, line 72: You missed something here

web/ui/dashboard/src/app/private/components/create-source/create-source.component.ts
```diff
@@ -85,7 +87,7 @@ export class CreateSourceComponent implements OnInit {
 		try {
 			const response = this.action === 'update' ? await this.createSourceService.updateSource({ data: sourceData, id: this.sourceId }) : await this.createSourceService.createSource({ sourceData });
 			this.isloading = false;
-			this.onAction.emit(response.data);
+			this.onAction.emit({ action: this.action, data: response.data });
```
horlah, line 90: This change here, please search everywhere we're using this component and reflect this change

horlah (approved):

PR #874, opened 2022-07-18
refactor server package

Dotunj:


server/server.go
```diff
@@ -0,0 +1,611 @@
+package server
```
Subomi, line 1: @Dotunj, I think we want to make this package re-usable as well. So that other services can use them too.

Like moving this guy into `internal/pkg/server`

Subomi: Secondly, we need to decouple the server from the router/handler. This is needed so that other services can bring their handlers. 

An API like:

```Golang
 
   apiHandler := buildRoutes()

   server := &Server{
      Handler: apiHandler,
   }
```
In the client code.

Subomi (changes requested):
@Dotunj FYI

internal/pkg/server/server.go
```diff
@@ -0,0 +1,92 @@
...
+type Server struct {
+	s *http.Server
+	M *middleware.Middleware
+}
+
+type CreateServer struct {
+	Cfg config.Configuration
+}
+
+func NewServer(c *CreateServer) *Server {
```
Subomi, line 29: We can refactor this further to use the exact dependency required: `func NewServer(port int) *Server`

Subomi (changes requested):

Subomi (approved):

PR #877, opened 2022-07-19
add support for switch command

Dotunj:


cmd/switch.go
```diff
@@ -0,0 +1,78 @@
...
+func addSwitchCommand() *cobra.Command {
+	var projectId string
+
+	cmd := &cobra.Command{
+		Use:               "switch",
+		Short:             "Switches the current project context using your Project ID",
+		PersistentPreRun:  func(cmd *cobra.Command, args []string) {},
+		PersistentPostRun: func(cmd *cobra.Command, args []string) {},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			homedir := os.Getenv("HOME")
```
jirevwe, line 24: I think we can use 
```suggestion
		homedir, err := os.UserHomeDir()
		if err != nil {
			return err
		}
```

jirevwe (approved):

PR #879, opened 2022-07-20
Retention Policy

Oluwadaminiola:
In this pr,
- the retention policy feature was added to the UI
- the organization settings were added to a new settings page
- Modifications were made to the settings and user settings menu


horlah:
Screenshot 2022-07-20 at 10 30 55 PM

- The line below the analytics toggle is missing here
- The input fields here have a weird margin-bottom with them, please review and remove
- The text below the analytics should be grey
- Should it be "s3" or "S3", is that how they describe it?
- On this settings page, make configuration the first option on the side navigation and selected option

@Oluwadaminiola

horlah:
For the margin-bottom on the input field, it is probably from the input component. Please review and consider removing it generally.

horlah (approved):

PR #882, opened 2022-07-20
fix: add latest tag to images

Dotunj:


.goreleaser.yml
```diff
@@ -38,7 +38,9 @@ publishers:
 
 
 dockers:
-  - image_templates: ["docker.cloudsmith.io/convoy/convoy/{{ .Env.REPO_NAME }}:{{ .Tag }}"]
+  - image_templates: 
+    - "docker.cloudsmith.io/convoy/convoy/{{ .Env.REPO_NAME }}:latest"
```
Subomi, line 42: @Dotunj With this will we have multiple `latest` or the most recent?

Subomi (commented):
@Dotunj With this, will we have multiple `latest` or only the most recent?

Dotunj:
@Subomi We'll have multiple `latest` with this.

Subomi:
@Dotunj, I think we should have one `latest`?

Dotunj:
@Subomi that works, however that means can we can only tag `latest` against the publish workflow i.e when we publish to Apt, Yum and HomeBrew.

Subomi:
That means release candidates won't get a latest tag?

Dotunj:
Yeah

Subomi:
@Dotunj, I can't remember exactly why we don't publish release candidates? Maybe we should? :)

Dotunj:
@Subomi I guess if someone is downloading and installing the latest version of a software, you want it to be the most recent stable version?

Subomi:
@Dotunj Hm, but we can't have multiple `latest`. The problem with our install scripts is when someone runs it and doesn't specify a version, it should pick the `latest` and install. 

Let us tag release candidates as `latest` for docker releases because I can set my version later, unlike apt, yum and brew, where I can't specify the version I want.

PR #892, opened 2022-07-25
fix:  force resend bug

Oluwadaminiola:
- the correct response while trying to resend/force resend an event is updated
- the arrow icons on the table and pagination buttons were updated


horlah:
@Oluwadaminiola Please check out the pagination arrow icon for the sources page too

horlah (approved):

PR #931, opened 2022-08-10
Extend cli functionality

jirevwe:
This PR extends the functionality of the client and server CLI to enable streaming events from a hosted convoy instance to their local machine without the use of a 3rd party tool. See more here #893


danvixent (approved):

services/client.go
```diff
@@ -0,0 +1,135 @@
...
+
+const (
+	// Maximum message size allowed from peer.
+	maxMessageSize = 512
+
+	maxDeviceLastSeenDuration = 2 * time.Minute
+)
+
+// Client is a middleman between the websocket connection and the hub.
+type Client struct {
```
Subomi, line 23: I think we can refactor this to be clearer:

- This type doesn't need a reference to the `hub`, the `hub` holds a reference to it already. 
- We can move this to `internal/pkg` and test it in isolation.

services/hub.go
```diff
@@ -0,0 +1,463 @@
...
+	Device *datastore.Device      `json:"device"`
+	Group  *datastore.Group       `json:"group"`
+	App    *datastore.Application `json:"app"`
+}
+
+type AckEventDelivery struct {
+	UID string `json:"uid"`
+}
+
+func (h *Hub) LoginHandler(w http.ResponseWriter, r *http.Request) {
```
Subomi, line 237: We can potentially make this better by:
1. Decoupling the websocket server http handlers away from the `hub`.
2. Using contexts to pass the `hub` down to the handlers where necessary.

services/client.go
```diff
@@ -0,0 +1,135 @@
...
+	hub *Hub
+
+	// device id of the cli client
+	deviceID   string
+	EventTypes []string
```
Subomi, line 28: We create subscriptions off the client's request. Is there any reason why we need `EventTypes` on this type?

services/hub_test.go
```diff
@@ -0,0 +1,721 @@
...
+			dbFn: func(h *Hub) {
+				d := h.deviceRepo.(*mocks.MockDeviceRepository)
+				d.EXPECT().FetchDeviceByID(gomock.Any(), "device-id", "abc", "1234").Times(1).Return(nil, errors.New("device not found"))
+			},
+			wantErr:     true,
+			wantErrCode: http.StatusBadRequest,
+			wantErrMsg:  "device not found",
+		},
+		{
+			name: "should_listen_successfully",
```
Subomi, line 190: Duplicate test case?

Subomi (changes requested):
Nice. @jirevwe 

I left some comments.

net/dispatcher.go
```diff
@@ -101,6 +103,58 @@ func (d *Dispatcher) SendRequest(endpoint, method string, jsonData json.RawMessa
 	return r, nil
 }
 
+func (d *Dispatcher) SendCliRequest(url string, method convoy.HttpMethod, apiKey string, jsonData json.RawMessage) (*Response, error) {
```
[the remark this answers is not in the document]
jirevwe: It doesn't need to merge the headers, we only use it for the login request

server/subscription.go
```diff
@@ -89,7 +89,7 @@ func (a *ApplicationHandler) CreateSubscription(w http.ResponseWriter, r *http.R
 		return
 	}
 
-	sub.Type = string(group.Type)
+	// sub.Type = string(group.Type)
```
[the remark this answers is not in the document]
jirevwe: Yep, the does not exist on the struct anymore

cmd/stream.go
```diff
@@ -0,0 +1,101 @@
...
+func addStreamCommand(a *app) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "stream",
+		Short: "Start a websocket server to pipe events to another convoy instance",
+		Run: func(cmd *cobra.Command, args []string) {
+			// enable only the native auth realm
+			authCfg := &config.AuthConfiguration{
+				Native: config.NativeRealmOptions{Enabled: true},
+			}
+			err := realm_chain.Init(authCfg, a.apiKeyRepo, nil, nil)
```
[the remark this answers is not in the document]
jirevwe: I don't understand what the issue is, should we move line 31 somewhere else?

util/crypto.go
```diff
@@ -44,6 +48,45 @@ func ComputeJSONHmac(hash, data, secret string, order bool) (string, error) {
 	return e, nil
 }
 
+type Signature struct {
```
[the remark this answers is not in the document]
jirevwe: Hmm, I don't think it should be moved. It depends on the util and the algo pkgs

cmd/listen.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+var done chan interface{}
+var interrupt chan os.Signal
+
+const (
+	// Time allowed to write a message to the peer.
+	writeWait = 10 * time.Second
```
oneEyedSunday, line 30: 10 seconds is a long time to wait to write something.

cmd/listen.go
```diff
@@ -0,0 +1,237 @@
...
+			}
+
+			listenRequest := socket.ListenRequest{
+				HostName:   c.Host,
+				DeviceID:   c.ActiveDeviceID,
+				SourceID:   source,
+				EventTypes: strings.Split(events, ","),
+			}
+
+			body, _ := json.Marshal(listenRequest)
```
oneEyedSunday, line 69: we are ignoring error from marshaling here, but go on to use / check it in the next line

internal/pkg/socket/client.go
```diff
@@ -0,0 +1,167 @@
...
+	err := c.conn.Close()
+	if err != nil {
+		log.WithError(err).Error("failed to close client conn")
+	}
+	unregister <- c
+}
+
+func (c *Client) GoOffline() {
+	c.lock.Lock()
+
```
oneEyedSunday, line 128: should defer the `c.lock.Unlock()` here

Subomi (changes requested):
@jirevwe FYI

Subomi (approved):

cmd/main.go
```diff
@@ -269,6 +272,10 @@ func preRun(app *app, db *cm.Client) func(cmd *cobra.Command, args []string) err
 
 func postRun(app *app, db *cm.Client) func(cmd *cobra.Command, args []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		if db == nil {
```
Dotunj, line 275: Do we still need this? I thought it was fixed already

server/models/models.go
```diff
@@ -244,8 +244,8 @@ type WebhookRequest struct {
 }
 
 type Subscription struct {
-	Name       string `json:"name" bson:"name" valid:"required~please provide a valid subscription name"`
-	Type       string `json:"type" bson:"type" valid:"required~please provide a valid subscription type"`
+	Name string `json:"name" bson:"name" valid:"required~please provide a valid subscription name"`
+	//Type       string `json:"type" bson:"type" valid:"required~please provide a valid subscription type"`
```
Dotunj, line 248: Are we taking this field out?

mocks/notification.go
```diff
@@ -0,0 +1,5 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: internal/notifications/notifications.go
+
+// Package mocks is a generated GoMock package.
+package mocks
```
Dotunj, line 5: I don't think this file is meant to exist.

server/models/models.go
jirevwe: Yeah, it's not being used. We don't need it since subscriptions are already scoped to a project which has the type already

mocks/notification.go
jirevwe: Oh, might have been added by mistake

danvixent (approved):

PR #937, opened 2022-08-11
Empty state story

Oluwadaminiola:


web/ui/dashboard/src/stories/empty-state/Empty-state.stories.ts
```diff
@@ -0,0 +1,43 @@
...
+		},
+        buttonText: {
+			control: { type: 'string' }
+		},
+		type: {
+			options: ['normal', 'table'],
+			control: { type: 'select' },
+			defaultValue: 'normal'
+		},
+        class: {
```
horlah, line 25: This is not needed here

horlah: We're meant to show the components in their natural state

horlah (changes requested):

horlah:
Screenshot 2022-08-16 at 3 03 50 PM

I can't edit controls here.

horlah:
I might have misdirected about removing the class property in the stories you created. Leaving the class there will allow us to test some styles on the component while testing or before applying it to the main component; since we're using Tailwind everywhere.

What do you think @Oluwadaminiola

Oluwadaminiola:
> I might have misdirected about removing the class property in the stories you created. Leaving the class there will allow us to test some styles on the component while testing or before applying it to the main component; since we're using Tailwind everywhere.
> 
> What do you think @Oluwadaminiola

yeah, I kind of agree with this, although due to the fact that most component stories are just standalone, the use cases for using the class property would be a bit limited, hence why I didn't necessarily add them to every story

PR #945, opened 2022-08-13
Refactor search to use scoped project collections

jirevwe:
This PR changes the way events are indexed for search. It removes the requirement for an extra `convoy index` command. Events will be indexed in the process event creation job handler and would be scoped to each project.


services/event_service.go
```diff
@@ -151,7 +152,36 @@ func (e *EventService) GetAppEvent(ctx context.Context, id string) (*datastore.E
...
+	filterByBuilder.WriteString(fmt.Sprintf(" && created_at:[%d..%d]", filter.SearchParams.CreatedAtStart, filter.SearchParams.CreatedAtEnd))
+
+	documents, paginationData, err := e.searcher.Search(filter.Group.UID, &datastore.SearchFilter{
+		Query:    filter.Query,
+		FilterBy: filterByBuilder.String(),
+		Pageable: filter.Pageable,
+	})
+
+	var ids []string
+	for _, d := range documents {
```
Subomi, line 173: @jirevwe, I think we should abstract this into the `typsense` package so it can be adequately tested as part of the package. I am thinking of the right abstraction, something like `e.searcher.SearchIndex` that returns an array of ids using the same internal search method, I believe.

worker/task/process_event_creation.go
```diff
@@ -154,6 +155,35 @@ func ProcessEventCreated(appRepo datastore.ApplicationRepository, eventRepo data
 			}
 		}
 
+		// convert event data field to map
```
Subomi, line 158: @jirevwe I think we should create another queue called `index-events` and create a separate task for it. It will be a low-priority queue that continues to index events from time to time. I think we will achieve better isolation to test that routine and take this out of our critical path of event delivery. What do you think? 

My first option was to see if `asynq` allows multiple handlers to read off the same queue, but It seems impossible.

worker/task/process_event_creation_test.go
```diff
@@ -131,80 +135,83 @@ func TestProcessEventCreated(t *testing.T) {
...
-
-				q, _ := args.eventQueue.(*mocks.MockQueuer)
-				q.EXPECT().Write(convoy.EventProcessor, convoy.EventQueue, gomock.Any()).Times(1).Return(nil)
+				srh, _ := args.search.(*mocks.MockSearcher)
+				srh.EXPECT().Index(gomock.Any(), gomock.Any())
 			},
 			wantErr: false,
 		},
+		// {
+		// 	name: "should_process_event_for_incoming_group",
```
Subomi, line 145: 👀

Subomi (changes requested):
@jirevwe Sweet. FYI

datastore/filter.go
```diff
@@ -14,3 +14,9 @@ type SourceFilter struct {
 	Type     string
 	Provider string
 }
+
+type SearchFilter struct {
+	Query    string
+	FilterBy   string
```
Subomi, line 20: @jirevwe By how much, we are doing `filterByBuilder` across it might be helpful to pack it into this type. I think we will have testing gains! ✊🏽

datastore/mongo/mongo.go
```diff
@@ -179,7 +179,6 @@ func (c *Client) ensureMongoIndices() {
 
 	c.ensureIndex(AppCollection, "group_id", false, nil)
 	c.ensureIndex(UserCollection, "uid", true, nil)
-	c.ensureIndex(UserCollection, "reset_password_token", true, nil)
```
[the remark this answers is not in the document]
jirevwe: Yes, the field shouldn't be in this collection tbh and it wasn't allowing the index process start

server/event.go
```diff
@@ -297,10 +297,6 @@ func (a *ApplicationHandler) GetEventsPaged(w http.ResponseWriter, r *http.Reque
 	query := r.URL.Query().Get("query")
 	app := r.URL.Query().Get("appId")
 
-	// if pageable.Page == 0 {
```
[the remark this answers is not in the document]
jirevwe: Yes, it was already commented out so I removed it

services/event_service.go
jirevwe: What you described was the original behavior, I just changed it in this PR because the package assumed you index an "event" when in reality you should be able to index any other collection

worker/task/process_event_creation.go
jirevwe: Sure

services/event_service.go
Subomi: No, but the structure of each document is the same, right? They all have a `uid`. We can filter for ids without assuming their collection.

worker/task/process_event_creation.go
jirevwe: I have modified this

datastore/filter.go
jirevwe: I don't understand

Subomi: We are doing that task repetitively. I think we can abstract it into that type and test it. Should we huddle?

Dotunj (approved):

Subomi (approved):
LGTM! @jirevwe

obbap1:
Hello @jirevwe, I see this error `Error: Get "http://localhost:8108/health": dial tcp 127.0.0.1:8108: connect: connection refused` when I run convoy and it looks tied to this commit and this struct specifically. 
```
Search: SearchConfiguration{
	Type: "typesense",
	Typesense: TypesenseConfiguration{
	        Host:   "http://localhost:8108",
	        ApiKey: "some-api-key",
	},
 },
```
Is there a flag/config i should be sending ?

jirevwe:
You'd need to setup a [typesense](https://github.com/typesense/typesense/) instance locally. If you don't specify a search backend config, you should not see that error since it'll default to using mongodb to search.

obbap1:
Okay, thank you

PR #953, opened 2022-08-19
add oss as default environment 

Dotunj:


jirevwe (approved):

config/config.go
```diff
@@ -159,9 +159,8 @@ type TypesenseConfiguration struct {
 }
 
 const (
-	envPrefix string = "convoy"
-
-	DevelopmentEnvironment string = "development"
+	envPrefix              string = "convoy"
+	OSSEnvironment         string = "oss"
```
Subomi, line 163: @Dotunj We should still have `DevelopmentEnvironment` for local dev, right?

Subomi (approved):

PR #981, opened 2022-08-29
fix endpoint notification email bug

### LINKED ISSUE #971: [BUG] Endpoint notification email is being sent multiple times
When re-activating an endpoint, instead of an email to be sent once, it is sent multiple times to the application's support email.

Dotunj:
Resolves #971


jirevwe (approved):

Subomi (changes requested):
@Dotunj Is it possible to write a test case to expose this bug?

Subomi (approved):

PR #988, opened 2022-08-31
fix middleware layer for app portal

Dotunj:


Subomi (changes requested):
@Dotunj Here as well, is it possible to write tests against the app portal routes to expose the bug?

Subomi (approved):

jirevwe (approved):

PR #992, opened 2022-09-06
updated event deliveries with cli icon

Oluwadaminiola:


jirevwe (approved):

horlah:
@Oluwadaminiola The identifier was added for event deliveries alone; doesn't the API doesn't return `device_id` for events?

PR #1003, opened 2022-09-09
Update subscriptions form

Oluwadaminiola:


web/ui/dashboard/src/app/private/components/create-subscription/create-subscription.component.html
```diff
@@ -62,7 +64,7 @@ <h2 class="font-semibold flex items-center mb-20px mt-30px">
 			<div class="grid grid-cols-2 gap-x-6" formGroupName="retry_config">
 				<convoy-select [options]="retryLogicTypes" name="type" formControlName="type" label="Mechanism" placeholder="Select mechanism"></convoy-select>
 
-				<convoy-input type="text" name="retry-logic-duration" formControlName="duration" label="Duration" placeholder="e.g 3s"></convoy-input>
+				<convoy-input type="text" name="retry-logic-duration" formControlName="duration" label="Duration" placeholder="e.g 3"></convoy-input>
```
horlah, line 67: I'm concerned about the change here. How will the user understand the unit to attribute to duration?
Will it be in minutes or seconds?

Can we use a tooltip to describe it or a more descriptive placeholder? @Oluwadaminiola

Oluwadaminiola: yes, I'll update this as there's still a slight change to be made

Dotunj (approved):

PR #1006, opened 2022-09-12
Refactor Store Dependency

Subomi:


datastore/mongo/application.go
```diff
@@ -154,24 +138,24 @@ func (db *appRepo) SearchApplicationsByGroupId(ctx context.Context, groupId stri
 	}
 
 	filter := bson.M{
-		"group_id":        groupId,
-		"document_status": datastore.ActiveDocumentStatus,
```
jirevwe, line 158: Was there a reason we are filtering out deleted groups any more?

datastore/mongo/event.go
```diff
@@ -164,18 +163,23 @@ func (db *eventRepo) FindEventByID(ctx context.Context, id string) (*datastore.E
 }
 
 func (db *eventRepo) FindEventsByIDs(ctx context.Context, ids []string) ([]datastore.Event, error) {
-	m := make([]datastore.Event, 0)
+	ctx = db.setCollectionInContext(ctx)
+
+	var event []datastore.Event
 
-	filter := bson.M{"uid": bson.M{"$in": ids}, "document_status": datastore.ActiveDocumentStatus}
+	filter := bson.M{"uid": bson.M{"$in": ids}}
```
jirevwe, line 170: If we remove the document status filter, won't it also return deleted records?

datastore/mongo/application.go
Subomi: @jirevwe I think it should be implicit. I made it the responsibility of the store. But there are several inconsistencies, some store methods add this scope others rely on the Caller to do set it. To achieve consistency, I think we should do it in another refactor PR.

datastore/mongo/event.go
Subomi: Yeah, similar issue.

danvixent (approved):

PR #1012, opened 2022-09-13
feat: add support for custom headers for outgoing events

### LINKED ISSUE #713: Allow a way to add a custom authorization headers to each payload sent to an application endpoint
If an application endpoint is behind an authorization system, the payload to be sent to the endpoint needs to include an Authentication header. This header should be configurable for each application endpoint because each endpoint will require a different credential.

This can be solved by adding "custom_headers" for each application endpoint. This property specifies the list of headers to be included on each payload and will be useful for other application also aside from authentication.

It should look something like this in the API:
```
"models.Endpoint": {
    "properties": {
	  ...
          "custom_headers": {
              "items": {
                  "key": {
                      "type":"string"
                  },
                  "value": {
                       "type":"string"
                  }
              },
              "type":"array"
          }
           
    },
    "type": "object"
}
```

Dotunj:
Resolves #713

### DIFF
--- a/datastore/models.go
+++ b/datastore/models.go
@@ -203,15 +203,33 @@ type Endpoint struct {
 	Description string `json:"description" bson:"description"`
 	Secret      string `json:"secret" bson:"secret"`
 
-	HttpTimeout       string `json:"http_timeout" bson:"http_timeout"`
-	RateLimit         int    `json:"rate_limit" bson:"rate_limit"`
-	RateLimitDuration string `json:"rate_limit_duration" bson:"rate_limit_duration"`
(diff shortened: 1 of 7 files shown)


Subomi (changes requested):
@Dotunj I think what we should be doing here is:

- Add a `custom_headers` field to `Event` type in `server/models.go`
- When processing requests from `POST /events`:
   - Retrieve headers in the `custom_headers` field.
   - Save them to the `Headers` field on the `Event` type in `datastore/models.go`

That should be all. The rest of the system should just work.

Subomi (approved):

jirevwe (approved):

PR #1030, opened 2022-09-20
add project apis to api/v1

danvixent:


server/security.go
```diff
@@ -128,6 +128,7 @@ func (a *ApplicationHandler) CreatePersonalAPIKey(w http.ResponseWriter, r *http
 // @Failure 400,401,500 {object} util.ServerResponse{data=Stub}
 // @Security ApiKeyAuth
 // @Router /ui/organisations/{orgID}/groups/{groupID}/apps/{appID}/keys [post]
+func _() {}
```
jirevwe, line 131: Why's this function empty?

jirevwe:
@danvixent @Subomi I think this is the best time to switch fully from `groupId` to `projectId` at least on the api level, or what do you guys think?

server/security.go
danvixent: For `swag` to generate docs for annotations, the annotations must precede a function declaration. This is actually a bug that was fixed in this pr. There were some annotations that never got generated previously.

danvixent:
@jirevwe aggre. This will come in another pr to the headless-convoy branch.

server/security.go
jirevwe: I see, ok

services/event_service.go
```diff
@@ -83,6 +85,11 @@ func (e *EventService) CreateAppEvent(ctx context.Context, newMessage *models.Ev
 		return nil, util.NewServiceError(http.StatusBadRequest, errors.New("app has no configured endpoints"))
 	}
 
+	// TODO(daniel): consider adding this check
+	// if app.GroupID != g.UID {
+	//	return nil, util.NewServiceError(http.StatusUnauthorized, errors.New("unauthorized to access app"))
+	// }
```
jirevwe, line 91: Why was this commented out?

jirevwe (approved):

services/event_service.go
danvixent: It's a TODO, maybe we should consider adding it cc @Subomi.
Because `app_id` is specified in the event request body we should verify that the app belongs to the group that the request was authenticated for.

internal/pkg/middleware/middleware.go
```diff
@@ -411,11 +404,13 @@ func (m *Middleware) RequireEvent() func(next http.Handler) http.Handler {
 
 func (m *Middleware) RequireOrganisation() func(next http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
-
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-
 			orgID := chi.URLParam(r, "orgID")
 
+			if util.IsStringEmpty(orgID) {
+				orgID = r.URL.Query().Get("orgID")
```
Dotunj, line 411: Is there any particular reason why we have to do this?

danvixent: There's no `orgID` path variable for the public apis, so it has to be sepcified via query.

Dotunj (approved):

PR #1051, opened 2022-10-04
Refactored Logging Implementation

Subomi:
**Todo**
- [x]  Write Unified Logging Package.
- [x]  Simplify Logging Configuration.
- [ ]  Create Dependency Injection for each component: Server, Consumer, Scheduler & Stream.
- [ ]  Create CLI Flags for each app component.


pkg/log/log.go
```diff
@@ -0,0 +1,192 @@
...
+func (l *Logger) WithLogger() *logrus.Logger {
+	return l.logger
+}
+
+// SetLevel sets the logger level.
+// It panics if v is less than DebugLevel or greater than FatalLevel.
+func (l *Logger) SetLevel(v Level) {
+	lvl, err := v.ToLogrusLevel()
+	if err != nil {
+		panic(err)
```
jirevwe, line 169: Should we panic here?

Subomi: I think yes, the reason being this method is called from the setup phase of the software. If we provide a level we can't log at, fail fast. There are 2 alternatives 

1. Return an error and allow the caller to panic. I think this is the same thing. 
2. Return an error and continue with a logger that logs at the default level, in that case, the user should not have supplied the log level flag in the first instance.

cmd/scheduler.go
```diff
@@ -4,36 +4,50 @@ import (
...
 			}
-			if cfg.Queue.Type != config.RedisQueueProvider {
-				log.WithError(err).Fatalf("Queue type error: Command is available for redis queue only.")
+
+			if util.IsStringEmpty(logLevel) {
+				logLevel = cfg.Logger.Level
+			}
+
+			lo := log.NewLogger(os.Stdout, "scheduler")
+			lvl, err := log.ParseLevel(logLevel)
```
Dotunj, line 40: Why don't we pass, parse and set the level when we call `NewLogger`?

cmd/scheduler.go
```diff
@@ -4,36 +4,50 @@ import (
...
+				return err
 			}
-			if cfg.Queue.Type != config.RedisQueueProvider {
-				log.WithError(err).Fatalf("Queue type error: Command is available for redis queue only.")
+
+			if util.IsStringEmpty(logLevel) {
+				logLevel = cfg.Logger.Level
+			}
+
+			lo := log.NewLogger(os.Stdout, "scheduler")
```
jirevwe, line 39: What's the difference between `a.logger` and this one?

cmd/scheduler.go
```diff
@@ -4,36 +4,50 @@ import (
...
 			cfg, err := config.Get()
 			if err != nil {
-				log.Fatalf("Error getting config: %v", err)
+				a.logger.Errorf("Failed to retrieve config: %v", err)
+				return err
 			}
-			if cfg.Queue.Type != config.RedisQueueProvider {
-				log.WithError(err).Fatalf("Queue type error: Command is available for redis queue only.")
+
+			if util.IsStringEmpty(logLevel) {
```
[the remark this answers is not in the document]
Subomi: Nice catch. I'll remove this check.

cmd/scheduler.go
Subomi: `a.logger` is used only by the `main` package. It is used only during app startup. This one is passed down and used by the `scheduler`.

jirevwe: Hmm, ok, but when starting up the process aren't we just running a single command? Is there a need to configure two different loggers?

Dotunj (approved):

jirevwe (approved):

PR #1053, opened 2022-10-06
Advanced Signatures Core Implementation

Subomi:


pkg/signature/signature.go
```diff
@@ -0,0 +1,171 @@
...
+	Encoding string
+}
+
+type Signature struct {
+	Payload json.RawMessage
+
+	// The order of this Schemes is a core part of this API.
+	// We use the index as the version number. That is:
+	// Index 0 = v0, Index 1 = v1
+	Schemes []Scheme
```
danvixent, line 46: @Subomi  don't you think we should allow explicit naming here ?

pkg/signature/signature.go
```diff
@@ -0,0 +1,171 @@
...
+		return nil, err
+	}
+
+	h := hmac.New(fn, []byte(secret))
+	h.Write(buf)
+
+	return h.Sum(nil), nil
+}
+
+func (s *Signature) getHashFunction(algo string) (func() hash.Hash, error) {
```
Dotunj, line 149: Are these the only hash algorithms we plan on supporting going forward? Because at the moment, we provide support for more than these two. @Subomi

internal/pkg/middleware/middleware.go
```diff
@@ -657,6 +657,7 @@ func (m *Middleware) RequireDeliveryAttempt() func(next http.Handler) http.Handl
 
 func (m *Middleware) findEndpoint(endpoints *[]datastore.Endpoint, id string) (*datastore.Endpoint, error) {
 	for _, endpoint := range *endpoints {
+		fmt.Printf("ENDPOINT: %+v\n", endpoint)
```
jirevwe, line 660: I've seen a few of these, we'd need to take them out before merging

internal/pkg/migrate/migrations.go
```diff
@@ -13,6 +16,167 @@ import (
...
+			}
+
+			for _, group := range groups {
+				if len(group.Config.Signature.Versions) == 0 {
+					continue
+				}
+
+				group.Config.Signature.Hash = group.Config.Signature.Versions[0].Hash
+				group.Config.Signature.Encoding = group.Config.Signature.Versions[0].Encoding.String()
+				group.Config.Signature.Versions = nil
```
jirevwe, line 77: Following up on the previous comment we set this one to nil correctly.

pkg/signature/signature.go
Subomi: @Dotunj The frontend will remove support for others except `SHA256` & `SHA512`

jirevwe (approved):

Dotunj (approved):

PR #1054, opened 2022-10-06
add support for feature flags

Dotunj:


jirevwe (approved):

internal/pkg/fflag/flipt/client.go
```diff
@@ -0,0 +1,85 @@
+package flipt
```
Subomi, line 1: @Dotunj This library isn't doing too much; we shouldn't be logging here. The client code should handle logging while this library returns errors.

cmd/main.go
```diff
@@ -112,6 +114,7 @@ type app struct {
 	cache    cache.Cache
 	limiter  limiter.RateLimiter
 	searcher searcher.Searcher
+	fflag    fflag.FeatureFlag
```
Subomi, line 117: @Dotunj We don't need this DI from the `cmd` package. From my understanding, this is only an HTTP Client. We should instantiate it where it's needed and call the API.

internal/pkg/fflag/noop/client.go
```diff
@@ -0,0 +1,11 @@
+package noopfflag
```
Subomi, line 1: @Dotunj I don't think we need a noop for this. From my understanding, It's an HTTP wrapper.

Subomi (changes requested):
Nice! Please see below:

jirevwe (approved):

danvixent (approved):

PR #1087, opened 2022-10-20
Fix: ui fixes

Oluwadaminiola:


web/ui/dashboard/src/app/private/components/create-subscription/create-subscription.component.ts
```diff
@@ -155,6 +155,11 @@ export class CreateSubscriptionComponent implements OnInit {
 		}
 	}
 
+	updateAppId(appId: string) {
```
horlah, line 158: Can't find where this method is called

horlah (changes requested):

horlah:
@Oluwadaminiola The select dropdown component should have a max height.

horlah (approved):

PR #1137, opened 2022-11-09
dropdown component rewrite

Oluwadaminiola:


horlah:
Nice work here @Oluwadaminiola.

Now we need to make this component such that the dropdown button and options are not predefined in the component, this component should simply be a container that enables dropdown UI. Something like this:
```
<convoy-dropdown>
    <button convoy-button dropdown-button>Normal Button</button>
    <ul convoy-dropdown-options>
        <li convoy-dropdown-option>Option Here</li>
    </ul>
</convoy-dropdown>
```

- convoy-dropdown is simply a container that creates the dropdown environment
- dropdown-button this should  be a directive that adds the CSS class names and event listener to the button that triggers the dropdown action
-  convoy-dropdown-options this should be a directive or component that creates the environment for the dropdown options
- convoy-dropdown-option this should be a directive that adds the event listeners for the dropdown option to close the dropdown on click.

PR #1143, opened 2022-11-11
update list component

Oluwadaminiola:


horlah:
One more thing here, we should be able to specify if we want a list to have a line or not. @Oluwadaminiola

horlah (approved):

PR #1160, opened 2022-11-17
add MaxIngestSize to group config

danvixent:


server/ingest.go
```diff
@@ -84,16 +84,27 @@ func (a *ApplicationHandler) IngestEvent(w http.ResponseWriter, r *http.Request)
 		}
 	}
 
-	cfg, err := config.Get()
+	g, err := createGroupService(a).FindGroupByID(r.Context(), source.GroupID)
```
Subomi, line 87: @danvixent Why can't we call the repository directly?

datastore/models.go
```diff
@@ -338,6 +338,7 @@ type GroupConfig struct {
 	Strategy                 *StrategyConfiguration        `json:"strategy"`
 	Signature                *SignatureConfiguration       `json:"signature"`
 	RetentionPolicy          *RetentionPolicyConfiguration `json:"retention_policy" bson:"retention_policy"`
+	MaxPayloadReadSize       uint64                        `json:"max_payload_read_size" bson:"max_payload_read_size"`
```
Subomi, line 341: @danvixent Let's change this to `MaxIngestSize`.

danvixent: @Subomi we also use it for response from endpoints after sending an event there. `MaxIngestSize` suggests it's only for ingesting events

worker/task/process_event_delivery.go
```diff
@@ -152,7 +152,12 @@ func ProcessEventDelivery(appRepo datastore.ApplicationRepository, eventDelivery
 		attemptStatus := false
 		start := time.Now()
 
-		resp, err := dispatch.SendRequest(e.TargetURL, string(convoy.HttpPost), sig.Payload, g, header, int64(cfg.MaxResponseSize), ed.Headers)
+		maxResponseSize := g.Config.MaxPayloadReadSize
```
Subomi, line 155: @danvixent This shouldn't change at all. We only want to limit our ingest size.

danvixent: alright

Subomi (approved):

jirevwe (approved):

PR #1167, opened 2022-11-18
update default retention policy to 30 days

danvixent:


jirevwe:
Why were docs generated for this PR? I might be but I don't think it should since the files in the server pkg were not modified.

danvixent:
There's a comment in Metadata struct `// NextSendTime denotes the next time an Event will be published in case it failed the first time` for some reason swag duplicate the comment everywhere.

jirevwe (approved):

PR #1180, opened 2022-11-24
remove annotations

Youngestdev:


server/organisation.go
```diff
@@ -52,19 +41,6 @@ func (a *ApplicationHandler) GetOrganisation(w http.ResponseWriter, r *http.Requ
 // @Router /api/v1/organisations [get]
 func _() {}
```
Subomi, line 42: @Youngestdev You can take this one out too.

Subomi, line 42: @Youngestdev, You can take this one out too.

Subomi (changes requested):

server/organisation.go
jirevwe: @danvixent isn't this the one you added, I can't remember why again

danvixent: @Subomi @Youngestdev please we shouldn’t remove those.

PR #1227, opened 2022-12-13
update email verification flow

Oluwadaminiola:


web/ui/dashboard/src/app/private/private.service.ts
```diff
@@ -256,4 +256,18 @@ export class PrivateService {
...
 		});
 	}
+
+    getUserDetails(requestDetails: { userId: string }): Promise<HTTP_RESPONSE> {
+		return new Promise(async (resolve, reject) => {
+			try {
+				const response = await this.http.request({
+					url: `/users/${requestDetails.userId}/profile`,
+					method: 'get'
+				});
```
horlah, line 266: We already have this endpoint call in account service; we shouldn't rewrite it here. @Oluwadaminiola

horlah (changes requested):

horlah (approved):

PR #1274, opened 2023-01-12
Feat: Use refresh tokens to keep users logged in

Oluwadaminiola:


web/ui/dashboard/src/app/services/http/http.service.ts
```diff
@@ -99,4 +95,40 @@ export class HttpService {
...
+			this.initiateLogout();
+		}
+	}
+
+	checkIfTokenIsExpired() {
+		const currentTime = new Date();
+		const tokenExpiryTime = this.jwtHelper.getTokenExpirationDate(this.authDetails().access_token);
+		if (tokenExpiryTime) {
+			const expiryPeriodInSeconds = differenceInSeconds(tokenExpiryTime, currentTime);
+			if (expiryPeriodInSeconds <= 180) this.getRefreshToken();
```
horlah, line 118: This might lead to a recursion @Oluwadaminiola. If the expiry period is less than 180s, you're making a request with the `request` method again to get the refresh token, which will again run the `checkIfTokenIsExpired` method.

Will this continue in circles, or am I reading it wrong?

horlah (changes requested):

web/ui/dashboard/src/app/services/http/http.service.ts
Oluwadaminiola: actually, because a refresh occurs whenever the `getRefreshToken()` method is called, the `expiryPeriodInSeconds` resets

Oluwadaminiola: so a recursion doesn't happen

horlah (approved):

PR #1285, opened 2023-01-17
adds support for pub/sub sources

### LINKED ISSUE #1009: [FEATURE REQUEST] Provide Support for Pub/Sub Sources
Allow different options for ingesting events asides HTTP. 

- [x] Add Support for  a new `convoy ingest` process
- [x] Add Support for SQS Sources
- [x] Add Support for Google Pub/Sub Sources

Dotunj:
Resolves #1009


internal/pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,239 @@
...
+type Source struct {
+	// The pub sub client.
+	client PubSub
+
+	// This is an identifier for the source config used to
+	// track if an existing source config has been changed.
+	hash string
+}
+
+type SourcePool struct {
```
Subomi, line 39: @Dotunj I think we can make this easier to understand - `SourcePool` handles the storage of sources. `SourceLoader` handles refreshing from the storage. 

I believe it'll make the APIs easier to reason about. The `SourcePool` will be a wrapper around maps for our own logic of `insertion`, `deletion`, `count` etc. At the same time, `SourceLoader` will be responsible for running at intervals to load new data from the backing store. 

What do you think?

Subomi: One example where it is confusing to reason about is two insert methods. :(

cmd/ingest.go
```diff
@@ -0,0 +1,55 @@
...
+var (
+	exit chan os.Signal
+)
+
+func addIngestCommand(a *app) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "ingest",
+		Short: "Ingest webhook events from Pub/Sub streams",
+		Run: func(cmd *cobra.Command, args []string) {
+			sourceRepo := mongo.NewSourceRepo(a.store)
```
Subomi, line 23: We are not using and passing down the `logger` here.

Subomi (changes requested):

Subomi (approved):

PR #1286, opened 2023-01-17
feat: regenerate project api key

Oluwadaminiola:


horlah:
@Oluwadaminiola We need a loading state for the generate button. Also, the button should be smaller, maybe like an underline text button or just a text button.

cc @Chloe-Ruth

horlah (approved):

PR #1310, opened 2023-01-30
implement postgres SourceRepository

danvixent:


database/postgres/source.go
```diff
@@ -0,0 +1,309 @@
...
+	countSources = `
+	SELECT COUNT(id) FROM convoy.sources WHERE deleted_at IS NULL;
+	`
+)
+
+var (
+	ErrSourceNotCreated         = errors.New("source could not be created")
+	ErrSourceVerifierNotCreated = errors.New("source verifier could not be created")
+	ErrSourceNotUpdated         = errors.New("source could not be updated")
+	ErrSourceNotRevoked         = errors.New("source could not be revoked")
```
jirevwe, line 110: I think this should be "deleted" and not "revoked"

jirevwe (approved):

PR #1394, opened 2023-02-24
Re-Implement subscription filters in Go

jirevwe:


pkg/compare/compare.go
```diff
@@ -0,0 +1,283 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+)
+
+type CompareFunc func(x, y interface{}) bool
+
+func defaultCompareMap() map[string]CompareFunc {
```
Subomi, line 11: @jirevwe Can this be a const? Can't think of a scenario where this will need to change.

jirevwe: Oh yeah

jirevwe: Hmm, apparently maps can't be made constants, I tried it and it didn't work.

Subomi: Maybe variable, then?

Subomi (approved):

PR #1397, opened 2023-02-24
replace mongo with postgres

Dotunj:


danvixent (approved):

surendrakaleyra:
what was the reason to migrate to postgres? or this could be done by adding it as a optional like people can choose whether they required mongo or postgres?

Issue #1400, opened 2023-02-26
dynamic query param support in endpoint


siprtcio:
Description:
We are using incoming webhook url with below configuration:

incoming webhook url
convoy.siprtc.io/ingest/1ihfJgwbeTDZNygn

endpoint url
http://tasks.apistack_cdrservice:8080/priv/Cdr/Data/

Now we are sending event like:
https://convoy.siprtc.io/ingest/1ihfJgwbeTDZNygn?uuid=c12b5bea-5c45-4431-98c7-5911ab2ce647

Our Expectation to get request url with query string on endpoint
http://tasks.apistack_cdrservice:8080/priv/Cdr/Data/?uuid=c12b5bea-5c45-4431-98c7-5911ab2ce647

but we are getting without query string
http://tasks.apistack_cdrservice:8080/priv/Cdr/Data/

Solution:
If dynamic replace variable can be supported in endpoint url then it will be great usecase

Subomi:
Hey @siprtcio , 

This use case makes sense. Thanks for submitting the feature request. We will ship support for this soon.

Subomi:
While you wait for this release, @siprtcio, you can use HTTP Headers to achieve this. We currently proxy down the headers to the endpoint.

subomi:
@danvixent Please can you port this #1404  to work for this

danvixent:
@subomi of course.

subomi:
Hey @vavido and @siprtcio This has been released in [v23.06.1](https://github.com/frain-dev/convoy/releases/tag/v23.06.1)

PR #1404, opened 2023-02-28
Queryparam support

surendratiwari3:
This PR will add extension to convoy for sending query param received on ingest url to endpoint url. This will also store it in event and event_delivery document. 

Referenced Issue

[https://github.com/frain-dev/convoy/issues/1400](url)

@Subomi


Subomi:
Hey @surendratiwari3, The build is failing. Please can you check it out?

surendratiwari3:
@Subomi can you check now?

Subomi:
@surendratiwari3 It seems the build still has problems.

services/event_service.go
```diff
@@ -410,15 +411,16 @@ func (e *EventService) createEvent(ctx context.Context, endpoints []datastore.En
...
-		CreatedAt: time.Now(),
-		UpdatedAt: time.Now(),
-		Endpoints: endpointIDs,
-		ProjectID: g.UID,
+		UID:         ulid.Make().String(),
+		EventType:   datastore.EventType(newMessage.EventType),
+		Data:        newMessage.Data,
+		Raw:         newMessage.Raw,
+		Headers:     e.getCustomHeaders(newMessage.CustomHeaders),
+		QueryParams: newMessage.QueryParams,
```
Subomi, line 419: @surendratiwari3 I think It's better to limit this to the ingest path alone. I can't think of a good use-case for doing this via the API. Also, the contract will be confusing as to what is for our REST API and what should be proxied.

datastore/models.go
```diff
@@ -571,6 +572,7 @@ type Event struct {
 	ProjectID        string                `json:"project_id,omitempty" db:"project_id"`
 	Endpoints        pq.StringArray        `json:"endpoints" db:"endpoints"`
 	Headers          httpheader.HTTPHeader `json:"headers" db:"headers"`
+	QueryParams      url.Values            `json:"query_params" bson:"query_params"`
```
Subomi, line 575: @surendratiwari3 We switched to `PostgreSQL` as our default store, you need to switch this up for it to work.

Subomi (changes requested):
@surendratiwari3 The build is passing now, but I left some feedback. :)

subomi:
Fixed in #1640

PR #1432, opened 2023-03-09
Update: UI Updates

Oluwadaminiola:
in this pr:
- the empty states were updated
- the create source form was updated with the source URL modal
- this Empty state, Modal, Card, and Page components were updated to directives
- Some minor UI updates were made


web/ui/dashboard/src/app/private/components/create-source/create-source.component.html
```diff
@@ -205,6 +205,14 @@ <h2 class="font-semibold mb-20px">Configure SQS</h2>
...
 </form>
 
+<convoy-token-modal
+	title="Source URL Generated"
+	description="Use this URL in your source platform to receive your webhook events."
+	notificationText="Source URL has been copied to your clipboard."
+	[token]="sourceData.url"
+	(closeModal)="showSourceUrl = false; onAction.emit({ action: this.action, data: sourceData })"
+	*ngIf="showSourceUrl"
+></convoy-token-modal>
```
horlah, line 215: Since this PR is a UI update, we should fix this modal thing once and for all. We should only have modal components as containers and only have something like this only when a specific usecase is reused more than twice. @Oluwadaminiola

web/ui/dashboard/src/app/private/components/create-source/create-source.component.ts
```diff
@@ -209,8 +211,9 @@ export class CreateSourceComponent implements OnInit {
 		this.isloading = true;
 		try {
 			const response = this.action === 'update' ? await this.createSourceService.updateSource({ data: sourceData, id: this.sourceId }) : await this.createSourceService.createSource({ sourceData });
-			this.onAction.emit({ action: this.action, data: response.data });
```
horlah, line 212: Why are we no longer emitting the status of this source form?
This is a create source component that has multiple use cases yea?

Or is there something I'm missing?

web/ui/dashboard/src/app/private/pages/project/event-logs/event-logs.component.html
```diff
@@ -118,11 +118,13 @@
 </div>
 
 <convoy-card>
-	<div class="flex">
+	<convoy-empty-state type="table" imgSrc="/assets/img/events-log-empty-state.png" className="min-h-[70vh]" heading="You currently do not have any event logs" *ngIf="!isloadingEvents && (!displayedEvents || displayedEvents?.length === 0)"></convoy-empty-state>
```
horlah, line 121: Empty state components, please lets rewrite them into a directive instead. Lets take out the use of `className` as much as possible

horlah (changes requested):

horlah:
@Oluwadaminiola Some of the empty state illustrations are repeated, it shouldn't be. The design should change.

web/ui/dashboard/src/app/components/modal/modal.component.html
```diff
@@ -1,37 +1,10 @@
...
-
-	<div class="m-auto empty:hidden" [class]="position === 'full' ? 'max-w-[834px]' : 'w-full p-20px'">
-		<ng-content select="[modalBody]"></ng-content>
-	</div>
-
-	<div class="fixed w-full bottom-0 px-20px pt-20px pb-16px border-t border-t-grey-10 empty:hidden">
-		<ng-content select="[modalFooter]"></ng-content>
-	</div>
+    <div convoy-modal-body [position]="position">
+        <ng-content select="[modalBody]"></ng-content>
```
horlah, line 6: I'm not sure we need to do this; think of it like this:
- Modal component is a container
- It knows all the possible shapes we might want our modal to have
- The only thing we want to pass into the ng-content when using the component is the modal title and body together, not separately. Why: there is no reason to have them separate.
```html
<div convoy-modal>
    <div convoy-modal-title></div>
    <div convoy-modal-body></div> // not sure if we'll need a directive like this though
</div>
```
- To close the modal, we should import an instance of the modal component and call a close modal method

web/ui/dashboard/src/app/components/modal/modal.component.ts
```diff
@@ -1,20 +1,48 @@
...
+})
+export class ModalHeaderComponent {
+	constructor() {}
+}
+
+
+@Directive({
+	selector: '[convoy-modal-body]',
+	standalone: true,
+	host: { class: 'm-auto empty:hidden', '[class]': "position === 'full' ? 'max-w-[834px]' : 'w-full p-20px'" }
```
horlah, line 27: We can set these values on the modal component, i.e. the width.

Hmm, except for the padding yea?

web/ui/dashboard/src/app/private/components/create-source/create-source.component.html
horlah: We're using token-modal for create source?

horlah (changes requested):

horlah (approved):

PR #1472, opened 2023-03-22
Change attempts field to bytea in postgres

danvixent:


jirevwe (changes requested):
The base branch should be dotun's branch

PR #1541, opened 2023-04-19
add support for meta events

### LINKED ISSUE #1020: [FEATURE REQUEST] Meta Events: Callback URL for Convoy's events.
**Describe Feature**
Convoy as a service should be able to publish webhook events about its webhook events.

**Why?** 
One use-case is when implementing the app portal, especially for users who have a dashboard already, not everyone wants to always have to call Convoy's API. They'd prefer to get notified about certain entities and show their users. Entities like: 
- Subscriptions Status
- EventDelivery Status

Dotunj:
Resolves #1020


internal/pkg/mevent/meta_event.go
```diff
@@ -0,0 +1,112 @@
+package mevent
```
Subomi, line 1: @Dotunj Let's move this to the `services` package. We would eventually move this to `internal/services`. We can rename it to `create_meta_event.go` as well.

jirevwe (approved):

Subomi (approved):

PR #1572, opened 2023-05-08
add support for ee release workflow

Dotunj:


.github/workflows/release-ee.yml
```diff
@@ -0,0 +1,75 @@
...
+on:
+    workflow_dispatch:
+        inputs:
+            name:
+                description: "Manual workflow name"
+                required: true
+    push:
+        tags:
+            # Release binary for every tag.
+            - v0.*
```
jirevwe, line 12: I don't think this tag will work anymore since we're not on 0.* anymore

scripts/ui.sh
```diff
@@ -0,0 +1,50 @@
...
+			else
+				helpFunc
+			fi
+
+			# Build Convoy
+			echo ""
+			buildUi
+			;;
+		?) helpFunc ;;
+		:) helpFunc ;;
```
jirevwe, line 43: Don't think this line is needed

scripts/ui.sh
```diff
@@ -0,0 +1,50 @@
...
+helpFunc() {
+	echo ""
+	echo "Usage: $0 -b"
+	echo -e "\t-b Specify the build flag (either ce or ee)"
+	exit 1
+}
+
+buildUi() {
+    # Enter UI directory
+    cd ./web/ui/dashboard
```
jirevwe, line 12: we need to exit incase this line fails
```suggestion
    cd ./web/ui/dashboard || exit 1
```

jirevwe (approved):

PR #1579, opened 2023-05-09
Separate dsn into parts

jirevwe:


api/server_suite_test.go
```diff
@@ -73,10 +73,10 @@ func getDB() database.Database {
 	return db
 }
 
-func getQueueOptions(name string) (queue.QueueOptions, error) {
+func getQueueOptions(_ string) (queue.QueueOptions, error) {
```
Dotunj, line 76: Is there a reason why we're not taking the argument out entirely?

cache/redis/client_test.go
```diff
@@ -17,7 +18,12 @@ type data struct {
 }
 
 func getDSN() string {
-	return os.Getenv("TEST_REDIS_DSN")
+	c := config.RedisConfiguration{
+		Scheme: "redis",
+		Host:   os.Getenv("TEST_REDIS_HOST"),
+		Port:   6379,
```
Dotunj, line 24: We're not using `TEST_REDIS_PORT` here.

subomi (approved):

Dotunj (approved):

PR #1592, opened 2023-05-17
Refactor security service

danvixent:


api/portal-api/portal_api.go
```diff
@@ -129,7 +129,7 @@ func (a *PortalLinkHandler) retrievePortalLink(r *http.Request) (*datastore.Port
 
 	if util.IsStringEmpty(token) {
 		cred, _ := middleware.GetAuthFromRequest(r)
-		token = cred.Token
+		token = cred.Token // TODO(all): this should probably be cred.APIKey, token is used for jwt see  middleware.GetAuthFromRequest(r)
```
jirevwe, line 132: Hmm, does this mean that we would need to pass the token in a different header?

services/create_api_key.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	dk := pbkdf2.Key([]byte(key), []byte(salt), 4096, 32, sha256.New)
+	encodedKey := base64.URLEncoding.EncodeToString(dk)
+
+	apiKey := &datastore.APIKey{
+		UID:       ulid.Make().String(),
+		MaskID:    maskID,
+		Name:      ss.NewApiKey.Name,
+		Type:      ss.NewApiKey.Type, // TODO: this should be set to datastore.ProjectKey
```
jirevwe, line 74: is this todo still valid?

jirevwe (changes requested):
It's fine, I just need info about those comments

Dotunj (approved):

api/portal-api/portal_api.go
danvixent: No, we found this fix.

services/create_api_key.go
danvixent: No, i've removed it

jirevwe (approved):

PR #1600, opened 2023-05-23
Refactor endpoint service

danvixent:


services/create_endpoint.go
```diff
@@ -0,0 +1,142 @@
...
+	Cache        cache.Cache
+	EndpointRepo datastore.EndpointRepository
+	ProjectRepo  datastore.ProjectRepository
+
+	E         models.Endpoint
+	ProjectID string
+}
+
+func (a *CreateEndpointService) Run(ctx context.Context) (*datastore.Endpoint, error) {
+	if err := util.Validate(a.E); err != nil {
```
subomi, line 29: @danvixent As a rule of thumb, model validations should happen in the handler.  Only other nuanced core business logic validations should happen here.

services/endpoint_service.go
```diff
@@ -1,23 +1,9 @@
 package services
```
subomi, line 1: @danvixent Do we still need this file?

subomi (changes requested):

services/endpoint_service.go
danvixent: @subomi I removed it

services/create_endpoint.go
danvixent: @subomi fixed

danvixent: @subomi fixed

services/endpoint_service.go
danvixent: @subomi I removed it

subomi (approved):

PR #1605, opened 2023-05-24
Add custom response to source

danvixent:


horlah:
@Oluwadaminiola You can remove the supposed link to the custom response documentation here, we don't have any documentation for it.

cc @subomi

subomi (changes requested):
@danvixent, @horlah FYI

1. Updating a source does not update the custom response body.
2. The update source form does not show the previously added response body but the previously added response content type.

subomi (approved):

qbx2:
Hello. Is it possible to template response using request header values? I want to solve okta's one-time verification request (https://developer.okta.com/docs/concepts/event-hooks/#one-time-verification-request) using convoy, but the challenge requires to respond with header value (x-okta-verification-challenge) .

PR #1616, opened 2023-05-29
feat: enhanced portal links

subomi:


datastore/models.go
```diff
@@ -1259,12 +1259,14 @@ type OrganisationInvite struct {
...
-	Endpoints         pq.StringArray   `json:"endpoints" db:"endpoints"`
-	EndpointsMetadata EndpointMetadata `json:"endpoints_metadata" db:"endpoints_metadata"`
+	UID                string           `json:"uid" db:"id"`
+	Name               string           `json:"name" db:"name"`
+	ProjectID          string           `json:"project_id" db:"project_id"`
+	Token              string           `json:"-" db:"token"`
+	OwnerID            string           `json:"owner_id" db:"owner_id"`
+	Endpoints          pq.StringArray   `json:"endpoints" db:"endpoints"`
+	EndpointsMetadata  EndpointMetadata `json:"endpoints_metadata" db:"endpoints_metadata"`
+	EndpointManagement bool             `json:"endpoint_management" db:"endpoint_management"`
```
jirevwe, line 1269: This is a weird name to give it

jirevwe, line 1269: This is a weird name to give it
```suggestion
	CanManageEndpoint bool             `json:"can_manage_endpoint" db:"endpoint_management"`
```

sql/1685202737.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
+ALTER TABLE convoy.portal_links
```
jirevwe, line 2: Do you think we'd need to add an index for the `owner_id` field? Seeing as we would be using it in queries

jirevwe (changes requested):
The endpoint_management field might need to be changed to be more readable, but the index needs to be added for the `owner_id field

Dotunj (approved):

PR #1622, opened 2023-05-30
refactor source request & response models

Dotunj:


api/models/source.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+}
+
+type SourceResponse struct {
+	*datastore.Source
+}
+
+// Generic function for looping over a slice of type M
+// and returning a slice of type T
+func NewListResponse[T, M any](items []M, fn func(item M) T) []T {
```
jirevwe, line 279: Should this function live here or in a general file?

jirevwe (approved):

api/models/source.go
Dotunj: Moved it to `api/models.go`

danvixent (approved):

PR #1624, opened 2023-05-31
Refactor subscription service

danvixent:


jirevwe (approved):

services/find_subscription_by_id.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, &ServiceError{ErrMsg: ErrSubscriptionNotFound.Error(), Err: err}
+	}
+
+	if s.SkipCache {
+		return sub, nil
+	}
+
+	// only incoming s.Projects have sources
+	if s.Project.Type == datastore.IncomingProject && sub.SourceID != "" {
+		source, err := s.SourceRepo.FindSourceByID(ctx, s.Project.UID, sub.SourceID)
```
Dotunj, line 35: We already fetch the source and endpoint details in the repo call for `FindSubscriptionByID`. Do we still need to do this here?

services/subscription_service.go
```diff
@@ -16,12 +16,8 @@ import (
 )
 
 var (
-	ErrSubscriptionNotFound               = errors.New("subscription not found")
 	ErrUpateSubscriptionError             = errors.New("failed to update subscription")
-	ErrCreateSubscriptionError            = errors.New("failed to create subscription")
 	ErrDeletedSubscriptionError           = errors.New("failed to delete subscription")
```
Dotunj, line 20: Do we still need this file?

danvixent: No, removed it.

PR #1629, opened 2023-06-01
Refactor event service

danvixent:


Makefile
```diff
@@ -17,7 +17,7 @@ build:
 
 integration_tests:
 	go run ./cmd migrate up
-	go test -tags integration -p 1 ./...
+	go test -tags integration -timeout 15m -p 1 ./...
```
jirevwe, line 20: remove this, we should not have to set an explicit timeout.

services/create_event.go
```diff
@@ -0,0 +1,166 @@
...
+	ErrNoValidOwnerIDEndpointFound = errors.New("owner ID has no configured endpoints")
+	ErrInvalidEndpointID           = errors.New("please provide an endpoint ID")
+)
+
+type CreateEventService struct {
+	EndpointRepo datastore.EndpointRepository
+	Queue        queue.Queuer
+
+	NewMessage *models.CreateEvent
+	G          *datastore.Project
```
jirevwe, line 32:
```suggestion
	P          *datastore.Project
```
Sholdn't this be `P`?

jirevwe (changes requested):
Remove the timeout from the test command

jirevwe (approved):

PR #1695, opened 2023-07-07
feat: enable multi-archtecture build for docker

nimboya:
Enable Multi-Architecture to build for convoy docker image

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -26,9 +26,18 @@ jobs:
         DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
         DOCKER_PASSWORD: ${{ secrets.PAT }}
         DOCKER_REGISTRY_URL: ghcr.io
+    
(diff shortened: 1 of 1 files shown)


jirevwe:
Hey @nimboya, apologies for the very delayed reply.

We currently use Goreleaser to build and publish images and that workflow file is a no-op as the workflow itself has been disabled for over a year now.

nimboya:
> Hey @nimboya, apologies for the very delayed reply.
> 
> We currently use Goreleaser to build and publish images and that workflow file is a no-op as the workflow itself has been disabled for over a year now.

Hi @jirevwe 

Does this mean docker images are not build anymore, if they are, is Goreleaser used to build and push it ?

jirevwe:
@nimboya They are being built, Goreleaser is used to build and publish them. See them [here](https://cloudsmith.io/~convoy/repos/convoy/packages)

subomi:
Opened a PR to do this with goreleaser here - #1790

PR #1699, opened 2023-07-12
fix: fix delete projects

Oluwadaminiola:


web/ui/dashboard/src/app/private/private.service.ts
```diff
@@ -378,6 +378,21 @@ export class PrivateService {
 		});
 	}
 
+    deleteProject(): Promise<HTTP_RESPONSE> {
```
horlah, line 381: @Oluwadaminiola Why did you move this here?

Oluwadaminiola: I wanted to deprecate the settings service file as it only contained only one service @horlah

horlah (approved):

PR #1700, opened 2023-07-12
Add support for Redis Clusters 

### LINKED ISSUE #1696: [CON-99] [Enhancement] Support Redis Cluster
Support Redis Cluster for increased high availability

<sub>[CON-99](https://linear.app/convoy-engineering/issue/CON-99/[enhancement]-support-redis-cluster)</sub>

jirevwe:
Fixes #1696

### DIFF
--- a/internal/pkg/rdb/rdb.go
+++ b/internal/pkg/rdb/rdb.go
@@ -10,32 +10,39 @@ import (
 
 // Redis is our wrapper logic to instrument redis calls
 type Redis struct {
-	dsn    string
-	client *redis.Client
+	addresses []string
+	client    redis.UniversalClient
 }
 
 // NewClient is used to create new Redis type. This type
 // encapsulates our interaction with redis and provides instrumentation with new relic.
-func NewClient(dsn string) (*Redis, error) {
-	if util.IsStringEmpty(dsn) {
-		return nil, errors.New("redis dsn cannot be empty")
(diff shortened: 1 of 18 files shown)


internal/pkg/rdb/rdb.go
```diff
@@ -2,40 +2,54 @@ package rdb
...
+
+		client = redis.NewClient(opts)
+		// Add Instrumentation
+		client.AddHook(nrredis.NewHook(opts))
+	} else {
+		client = redis.NewUniversalClient(&redis.UniversalOptions{
+			Addrs: addresses,
+		})
+		// Add Instrumentation
+		client.AddHook(nrredis.NewHook(nil))
```
subomi, line 45: @jirevwe Will this retain instrumentation for the cluster?

subomi (approved):

internal/pkg/rdb/rdb.go
jirevwe: It should, though it'll have to be tested to confirm it. The param is the options which is the redis dsn in parts we don't need that for a redis cluster as the API doesn't return it

jirevwe: Test it, it's retained

subomi: Make sense!

danvixent (approved):

PR #1714, opened 2023-07-25
feat: show/hide signup button on login page

Oluwadaminiola:


horlah:
1. There should be a loader for this API, and handle network error state.
2. Having this on the login page is not enough, every page that might lead to the signup page and the signup page itself should be connected to a service that manages the signup state.

@Oluwadaminiola

web/ui/dashboard/src/app/guards/signup/signup.guard.ts
```diff
@@ -0,0 +1,21 @@
...
+import { LoginService } from 'src/app/public/login/login.service';
+
+@Injectable({
+	providedIn: 'root'
+})
+export class SignupGuard implements CanActivate {
+	constructor(private router: Router, private loginService: LoginService) {}
+
+	canActivate() {
+		const isSignupEnabled = this.loginService.signupConfig && location.hostname !== 'dashboard.getconvoy.io';
```
horlah, line 12: This isn't just for our cloud @Oluwadaminiola

horlah: In any instance, nobody should be able to use the signup page if signup is disabled in Convoy config.

web/ui/dashboard/src/app/public/login/login.service.ts
```diff
@@ -30,10 +30,21 @@ export class LoginService {
 					url: '/configuration/is_signup_enabled',
 					method: 'get'
 				});
+
+				localStorage.setItem('IS_SIGNUP_ENABLED', JSON.stringify(response.data));
```
horlah, line 34: No no, we should not use localstorage, state in the service should be enough.

horlah (changes requested):

web/ui/dashboard/src/app/public/login/login.service.ts
```diff
@@ -22,4 +22,29 @@ export class LoginService {
...
+				return resolve(response);
+			} catch (error) {
+				return reject(error);
+			}
+		});
+	}
+
+	get signupConfig() {
```
horlah, line 42: Why do we need this?

horlah (changes requested):

web/ui/dashboard/src/app/public/login/login.service.ts
Oluwadaminiola: this is what I'm calling in the signup guard to check if signup is enabled

web/ui/dashboard/src/app/app-routing.module.ts
```diff
@@ -24,6 +27,7 @@ const routes: Routes = [
 	},
 	{
 		path: '',
+		resolve: [() => getSignupConfig()],
```
horlah, line 30: I don't think this is the right place for this. Actually, I don't think we should have it at all.

This guy only has business with the login and signup page, let those two pages handle it, along with a loader state.

horlah (changes requested):
One last thing

web/ui/dashboard/src/app/app-routing.module.ts
Oluwadaminiola: I added this here to be able to get the signup config value whenever a public route is being accessed, calling this only at the login page would restrict fetching the signup config value to only when the login page is being accessed, we can't fetch this value on the signup page as this page shouldn't even be accessed at all if the config value is being returned as `false`, hence why this is here, on any instance, a public route is always the first route accessed before any other activity is carried out, hence why that function is being called there, so the signup guard would have an accurate value to work with at all times because the config has already been fetched.

horlah: The signup guard does this exact same thing

PR #1726, opened 2023-08-09
Refactor event delivery process to optimize resource usage

jirevwe:


danvixent (approved):

util/msgpack.go
```diff
@@ -0,0 +1,34 @@
+package util
```
subomi, line 1: I think this file should be in `/internal/pkg`.

subomi (changes requested):

util/msgpack.go
jirevwe: Why? It doesn't import any convoy specific pkgs

subomi: Oh, yes, then it should be `/pkg` directly, rather than `util`.

subomi (approved):

PR #1744, opened 2023-08-24
fix: pause events polling on search and filtering

Oluwadaminiola:
In this pr, an update was made to pause events and event deliveries whenever a search or filtering is initiated, this is to help minimize data conflict whenever a search or filter request is made.


horlah:
@Oluwadaminiola The polling does not resume after applying a filter.

jirevwe (approved):

PR #1764, opened 2023-09-19
Performance tweaks

caplonso:
Hello

Last week our team ran a load test on convoy and noticed that the worker module was consuming a lot of cache memory and database CPU. So we started investigating the root cause and identified some points that would benefit from caching and also that the limitRater class could be passed as an argument to the event delivery function. The changes mentioned provided us with a gain of approximately 300% in application performance and also reduced resource usage. Below is a list of the most relevant changes we made:

  - 2 minute cache of subscription configuration in event creation consumer
  - Creation of indexing events only if the search engine is typesense
  - Creation of an environment variable that allows controlling the size of the connection pool for asynq consumers
  - Early return in the event processing function if the event status is success
  - Removal of the left join with the events_endpoints table when fetching event endpoint data.
 
We also made an adjustment to the function that searches events by status that is used in dashboard and now it is possible to use it in searches.

I hope you find changes that are satisfactory and that can be used in your project.

PS: We have not reviewed existing tests and they are likely to fail

### DIFF
--- a/api/dashboard/event.go
+++ b/api/dashboard/event.go
@@ -409,8 +409,26 @@ func (a *DashboardHandler) GetEventsPaged(w http.ResponseWriter, r *http.Request
 		pagedResponse{Content: &m, Pagination: &paginationData}, http.StatusOK))
 }
 
+func parseStatus(rawStatus string) []datastore.EventDeliveryStatus {
+	var parsedStatus []string
+	err := json.Unmarshal([]byte(rawStatus), &parsedStatus)
+
+	if err != nil {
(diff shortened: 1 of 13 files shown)


jirevwe:
Hey @caplonso,

Thanks for taking the time to open this PR. I have gone through it and I like what I'm seeing. Might I ask what version you built these off, I can see code from some months ago which we've removed.

caplonso:
Hi @jirevwe 

We used the version in the tag v23.05.5 which is the same used in our cluster currently.

Regards!

PR #1783, opened 2023-10-03
Performance tweaks (#1765)

jirevwe:
This PR
- Adds a write-through cache which speeds up reads
- Removes the dependency on Typsense for search
- Removed all db calls in the main ingress event handler


api/public/public.go
```diff
@@ -137,22 +136,29 @@ func (a *PublicHandler) BuildRoutes() http.Handler {
...
 func (a *PublicHandler) retrieveProject(r *http.Request) (*datastore.Project, error) {
 	projectID := chi.URLParam(r, "projectID")
 
 	if util.IsStringEmpty(projectID) {
-		return &datastore.Project{}, errors.New("Project ID not present in request")
+		return nil, errors.New("project id not present in request")
+	}
+
+	var project *datastore.Project
+	// fetch project from context or cache
```
danvixent, line 154: Comment no longer relevant

configs/local/docker-compose.yml
```diff
@@ -1,88 +1,84 @@
...
+        image: rtukpe/bench:v2
+        command: ["./cmd", "server", "--config", "convoy.json", "--port", "3001"]
+        volumes:
+            - ./convoy.json:/convoy.json
+        restart: on-failure
+        networks:
+            backendCluster:
+                ipv4_address: 172.16.0.11
+        ports:
+            - "3001:3001"
```
danvixent, line 27: Does this need to be here or a separate compose file?

database/postgres/endpoint.go
```diff
@@ -331,12 +369,19 @@ func (e *endpointRepo) CountProjectEndpoints(ctx context.Context, projectID stri
 }
 
 func (e *endpointRepo) FindEndpointByTargetURL(ctx context.Context, projectID string, targetURL string) (*datastore.Endpoint, error) {
-	endpoint := &datastore.Endpoint{}
-	err := e.db.QueryRowxContext(ctx, fetchEndpointByTargetURL, targetURL, projectID).StructScan(endpoint)
-	if err != nil {
-		if errors.Is(err, sql.ErrNoRows) {
-			return nil, datastore.ErrEndpointNotFound
+	endpoint, err := e.readFromCache(ctx, targetURL, func() (*datastore.Endpoint, error) {
```
danvixent, line 372: Will targetURL work here?

subomi: @danvixent It should work, it's merely a cache key.

subomi (changes requested):

configs/local/docker-compose.yml
jirevwe: I'm reverting the compose file changes, I left it because I might still need to make changes

danvixent: Okay

api/dashboard/dashboard.go
```diff
@@ -176,19 +176,25 @@ func (a *DashboardHandler) retrieveMembership(r *http.Request) (*datastore.Organ
...
 	if util.IsStringEmpty(projectID) {
-		return &datastore.Project{}, errors.New("Project ID not present in request")
+		return nil, errors.New("project id not present in request")
 	}
 
-	projectRepo := postgres.NewProjectRepo(a.A.DB)
-	return projectRepo.FetchProjectByID(r.Context(), projectID)
+	var project *datastore.Project
+	projectRepo := postgres.NewProjectRepo(a.A.DB, a.A.Cache)
+	project, err := projectRepo.FetchProjectByID(r.Context(), projectID)
```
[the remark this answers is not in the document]
jirevwe: I see what you mean. I was formerly using the cache in that block

cmd/stream/stream.go
```diff
@@ -114,7 +114,8 @@ func AddStreamCommand(a *cli.App) *cobra.Command {
 			}
 			q := redisQueue.NewQueue(opts)
 
-			consumer := worker.NewConsumer(q, lo)
+			consumer := worker.NewConsumer(0, q, lo)
```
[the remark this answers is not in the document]
jirevwe: Yeah, I don't see the need though, the defaults will work for it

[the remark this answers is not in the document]
subomi: Ok. Then it's not quite clear tbh, you can leave a comment to clarify. At face value it looks like a bug, not knowing that `0` is a `nil` value that translates to max CPU.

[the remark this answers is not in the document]
jirevwe: Hmm true

database/postgres/endpoint.go
subomi: @danvixent, It's doing lazy loading; after a miss, it retrieves it from the DB and saves it using the `targetURL` as the key.

internal/pkg/pubsub/source_loader.go
```diff
@@ -125,73 +123,41 @@ func (s *SourceLoader) fetchProjectSources(ctx context.Context) error {
 	return nil
 }
 
-func (s *SourceLoader) handler(ctx context.Context, source *datastore.Source, msg string) error {
-	txn, innerCtx := apm.StartTransaction(ctx, fmt.Sprintf("%v handler", source.Name))
```
subomi, line 129: @jirevwe Any reason why we removed this?

subomi (changes requested):

api/models/event.go
```diff
@@ -11,13 +11,12 @@ import (
...
 type CreateEvent struct {
-	EndpointID string `json:"endpoint_id"`
-	AppID      string `json:"app_id"` // Deprecated but necessary for backward compatibility
-	EventType  string `json:"event_type" valid:"required~please provide an event type"`
-
-	// Data is an arbitrary JSON value that gets sent as the body of the
-	// webhook to the endpoints
+	UID            string
+	AppID          string            `json:"app_id"` // Deprecated but necessary for backward compatibility
+	OwnerID        string            `json:"owner_id"`
```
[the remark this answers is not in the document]
jirevwe: The UID is needed because we need to generate the ID in the handler so we can use it as the job on asynq.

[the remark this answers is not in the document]
jirevwe: I'll remove the `OwnerID` field.

internal/pkg/pubsub/source_loader.go
jirevwe: That was a mistake, I'll add it back

subomi (approved):

api/ingest.go
```diff
@@ -91,15 +99,11 @@ func (a *ApplicationHandler) IngestEvent(w http.ResponseWriter, r *http.Request)
...
 	}
 
-	projectRepo := postgres.NewProjectRepo(a.A.DB)
-
-	g, err := projectRepo.FetchProjectByID(r.Context(), source.ProjectID)
-	if err != nil {
-		_ = render.Render(w, r, util.NewServiceErrResponse(err))
-		return
+	var maxIngestSize uint64
+	if project.Config != nil && project.Config.MaxIngestSize == 0 {
```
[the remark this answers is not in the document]
jirevwe: Yeah, I'll update it

danvixent (approved):

PR #1804, opened 2023-10-16
fix: fixing the permission denied when create a volume in current dir

vitorscassiano:
## Context:
We are stucking when running the entire docker compose manifest in localhost, we've got the following message:

```shell
➜  local git:(main) docker compose logs postgres redis_server typesense
postgres-1 |chown: /data/postgres: Permission denied
redis_server-1 |chown: /data/postgres: Permission denied
typesense-1 |chown: /data/postgres: Permission denied
```

## Solution:
I propose split up the volumes from each component to volumes session, ex:
```yaml
volumes:
    postgres_data:
    typesense_data:
    redis_data:
```

## Reference:
- https://docs.docker.com/storage/volumes/#use-a-volume-with-docker-compose

### DIFF
--- a/configs/local/docker-compose.yml
+++ b/configs/local/docker-compose.yml
@@ -69,13 +69,13 @@ services:
           POSTGRES_PASSWORD: convoy
           PGDATA: /data/postgres
         volumes:
-          - ./postgres_data:/data/postgres
+          - postgres_data:/data/postgres
 
     redis_server:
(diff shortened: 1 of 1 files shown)


subomi:
Hey, @vitorscassiano, thanks for the pull request. Please check the CLA so I can merge this.

subomi (approved):

PR #1806, opened 2023-10-18
Optimise source loader query

danvixent:
Fixes CON-187


jirevwe (approved):

internal/pkg/pubsub/source_loader.go
```diff
@@ -89,18 +85,20 @@ func (s *SourceLoader) fetchSources(ctx context.Context, projectID string, curso
 	}
 
 	for _, source := range sources {
-		go func(source datastore.Source) {
-			ps, err := NewPubSubSource(&source, s.handler, s.log)
-			if err != nil {
-				s.log.WithError(err).Error("failed to create pub sub source")
-			}
+		ps, err := NewPubSubSource(&source, s.handler, s.log)
```
subomi, line 88: @danvixent I fixed this bug in a previous PR. The gorountine needs to be there to scope the source, or else we run into a concurrency bug. Check #1800

subomi (approved):

PR #1812, opened 2023-10-23
Add Prefix config for S3 storage

### LINKED ISSUE #1787: [CON-169] [Feature Request] Add prefix level partitioning for object storage.
**Description**

For multi-tenancy with S3, we need to support object key prefixes instead of the per-bucket model because AWS has a very limited quota for buckets per account. The default is 100, and there’s a hard stop at 1000. The preferred approach is to use prefixes—e.g. `tenant-1/object`, `tenant-2/object,` etc. You can read more [here](https://aws.amazon.com/blogs/apn/partitioning-and-isolating-multi-tenant-saas-data-with-amazon-s3/).

<sub>From [SyncLinear.com](https://synclinear.com) | [CON-169](https://linear.app/convoy-engineering/issue/CON-169/[feature-request]-add-prefix-level-partitioning-for-object-storage)</sub>

danvixent:
fixes #1787

### DIFF
--- a/database/postgres/configuration_test.go
+++ b/database/postgres/configuration_test.go
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"gopkg.in/guregu/null.v4"
+
 	"github.com/oklog/ulid/v2"
 
 	"github.com/frain-dev/convoy/datastore"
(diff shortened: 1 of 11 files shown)


database/postgres/configuration_test.go
```diff
@@ -93,6 +95,20 @@ func generateConfig() *datastore.Configuration {
 		UID:                ulid.Make().String(),
 		IsAnalyticsEnabled: true,
 		IsSignupEnabled:    false,
-		StoragePolicy:      &datastore.DefaultStoragePolicy,
```
jirevwe, line 96: I think we can delete the default storage policy

jirevwe (approved):

database/postgres/configuration_test.go
danvixent: We still use it in CreateConfigService

subomi (approved):

PR #1829, opened 2023-10-30
Update convoy.json.example

danvixent:


convoy.json.example
```diff
@@ -1,81 +1,115 @@
...
-      "ssl_key_file": "",
-      "port": 5005
+  "host": "<insert-host>",
+  "custom_domain_suffix": "<insert-custom-domain-suffix>",
+  "search": {
+    "type": "typesense",
+    "typesense": {
+      "host": "<insert-typesense-host>",
+      "api_key": "<insert-typesense-api-key>"
     }
```
jirevwe, line 93: Don't think this is needed

convoy.json.example
```diff
@@ -1,81 +1,115 @@
...
+  "auth": {
+    "file": {
+      "basic": {
+        "username": "<insert-username>",
+        "password": "<insert-password>"
+      },
+      "api_key": {
+        "api_key": "<insert-api-key>"
+      }
+    },
```
jirevwe, line 11: Are we still using file realm?

jirevwe (approved):

PR #1870, opened 2024-01-12
Endpoint api improvements

danvixent:


api/public_integration_test.go
```diff
@@ -862,8 +769,8 @@ func (s *PublicEventIntegrationTestSuite) Test_CreateEndpointEvent_With_App_ID_V
 	//var event datastore.Event
 	//parseResponse(s.T(), w.Result(), &event)
 
-	//require.NotEmpty(s.T(), event.UID)
-	//require.Equal(s.T(), event.Endpoints[0], endpointID)
+	// require.NotEmpty(s.T(), event.UID)
+	// require.Equal(s.T(), event.Endpoints[0], endpointID)
```
jirevwe, line 773: We should remove these lines and the ones before them. Is there a reason we're retaining them?

jirevwe (approved):

PR #1876, opened 2024-01-19
Fix: fix source and subscription forms

Oluwadaminiola:
In this PR, the following bugs were fixed:
- Retry config not sent in the subscription form
- Inability to update outgoing subscription form if retry config does not exist
The source form validation was updated as well.


web/ui/dashboard/src/app/private/components/create-source/create-source.component.ts
```diff
@@ -251,6 +251,9 @@ export class CreateSourceComponent implements OnInit {
 			this.isloading = false;
 			return this.sourceForm.markAllAsTouched();
 		}
+
+		// if (!this.addKafkaAuthentication) delete sourceData.pub_sub.kafka.auth;
+
```
horlah, line 256: Do still need to keep it?

subomi (approved):

PR #1878, opened 2024-01-20
PortalLink Improvements

### LINKED ISSUE #1881: [CON-279] [Bug report]: PortalLinks with owner id don't properly show the list of endpoints
<sub>From [SyncLinear.com](https://synclinear.com) | [CON-279](https://linear.app/convoy-engineering/issue/CON-279/[bug-report]-portallinks-with-owner-id-dont-properly-show-the-list-of)</sub>

jirevwe:
Fixes #1881

### DIFF
--- a/api/handlers/portal_link.go
+++ b/api/handlers/portal_link.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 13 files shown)


api/handlers/portal_link.go
```diff
@@ -247,7 +248,8 @@ func (h *Handler) LoadPortalLinksPaged(w http.ResponseWriter, r *http.Request) {
 	}
 
 	endpointIDs := getEndpointIDs(r)
-	filter := &datastore.FilterBy{EndpointIDs: endpointIDs}
+	ownerId := r.URL.Query().Get("ownerId")
```
subomi, line 251: @jirevwe In fact, this is all we need to solve the bug. 

```go
var portalLinkEndpoints []string 
endpointRepo := postgres.NewEndpointRepo(h.A.DB, h.A.Cache)
endpoints, err := endpointRepo.FindEndpointsByOwnerID(r.Context(), pl.ProjectID, ownerId)
if err != nil {
   return nil, err
}

for _, endpoint := range endpoints {
   portalLinkEndpoints = append(endpointIDs, endpoint.UID)
}
```

This feels weird, but that means you can retrieve portal links with both ownerId and endpointIds at the same time except we want to restrict that behavior.

subomi (changes requested):

api/handlers/portal_link.go
subomi: @jirevwe If we do this, we can solve the bug temporarily without the other changes.

PR #1854, opened 2023-12-19
Compress and remove export

slobodangacesa:
Compress exported files before uploading to S3 and remove them afrer upload

PR #1882, opened 2024-01-23
fix: refactor export worker implementation

subomi:
### What does this do
Replaces #1854. I refactored and implemented removing temp files once export is complete.


subomi:
@slobodangacesa, fyi, compression will be included in our pro plan, starting at $99/month for self-hosted instances.

jirevwe:
I can't see the compression code in this. Can you point me to it?

danvixent (approved):

PR #1884, opened 2024-01-23
Dynamic event api improvements

danvixent:


api/models/event.go
```diff
@@ -26,9 +27,13 @@ func (e *CreateEvent) Validate() error {
...
 
 type DynamicEvent struct {
-	Endpoint     DynamicEndpoint     `json:"endpoint"`
-	Subscription DynamicSubscription `json:"subscription"`
-	Event        DynamicEventStub    `json:"event"`
+	URL            string          `json:"url" valid:"required~please provide a url"`
+	Secret         string          `json:"secret" valid:"required~please provide a secret"`
+	EventTypes     []string        `json:"event_types"`
+	Data           json.RawMessage `json:"data" valid:"required~please provide your data"`
+	ProjectID      string          `json:"project_id" swaggerignore:"true"`
```
subomi, line 34: @danvixent, Why do we need the `ProjectID` in the payload? The API Key already holds the context.

worker/task/process_dynamic_event_creation.go
```diff
@@ -205,136 +201,45 @@ func ProcessDynamicEventCreation(endpointRepo datastore.EndpointRepository, even
...
-			SupportEmail:       newEndpoint.SupportEmail,
-			SlackWebhookURL:    newEndpoint.SlackWebhookURL,
-			TargetURL:          newEndpoint.URL,
-			Description:        newEndpoint.Description,
-			RateLimit:          newEndpoint.RateLimit,
-			HttpTimeout:        newEndpoint.HttpTimeout,
-			AdvancedSignatures: newEndpoint.AdvancedSignatures,
-			AppID:              newEndpoint.AppID,
-			RateLimitDuration:  duration.String(),
+			Title:              fmt.Sprintf("endpoint-%s", uuid.NewString()),
```
subomi, line 215: @danvixent Why not we attach it to the endpoint's UID so it's easy to match?

subomi (changes requested):
@danvixent FYI

subomi (approved):

PR #1887, opened 2024-01-25
Organisation changes

jirevwe:


danvixent (approved):

sql/1706098676.sql
```diff
@@ -0,0 +1,82 @@
+-- from https://github.com/geckoboard/pgulid
```
subomi, line 1: @jirevwe While this gives us transactional benefits. I wonder if it's still the right path to go. My issue is if we ever decide to change this logic. For example, we no longer want to auto-provision users/organisations like in the cloud but want to rely on the `bootstrap` command only for bootstrapping sake. It'd take a lot of work to implement. We'd need to write another migration to handle it or write code to remove the hardcoded user.

If we're trying to solve the lack of transactions in the `datastore` API only, then that can be solved at that layer. In fact, I plan to include it in the refactoring I'm doing.

subomi: Another option is to only add the functions and allow them take arguments that the `bootstrap` command can use.

PR #1897, opened 2024-01-30
License implementation

jirevwe:


api/handlers/portal_link.go
```diff
@@ -31,6 +33,12 @@ import (
 //	@Security		ApiKeyAuth
 //	@Router			/v1/projects/{projectID}/portal-links [post]
 func (h *Handler) CreatePortalLink(w http.ResponseWriter, r *http.Request) {
+	valid := license.LICENSE.Load()
```
subomi, line 36: @jirevwe, this would make sense as a middleware.

PR #1911, opened 2024-02-04
Feature - Add support for RabbitMQ source

nitzangoldfeder:
Hello Team

I've added support for RabbitMQ source for outgoing projects

The UI will enable to create a new pubsub called RabbitMQ / AMQP
Its required to provide shcema / host / port and queue name
- Optional: add auth, add queue bindings

On the backend:
- Added amqp client go internal/pkg/pubsub

Tested creation of this source on my local machine and validated that messages are being consumed on the backend


internal/pkg/pubsub/amqp/client.go
```diff
@@ -0,0 +1,141 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/frain-dev/convoy/datastore"
+	"github.com/frain-dev/convoy/pkg/log"
+	amqp "github.com/rabbitmq/amqp091-go"
+)
+
+var ErrInvalidCredentials = errors.New("your kafka credentials are invalid. please verify you're providing the correct credentials")
```
subomi, line 13: @nitzangoldfeder, we can remove this.

internal/pkg/pubsub/amqp/client.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if err != nil {
+		log.WithError(err).Error("failed to declare queue")
+		return
+	}
+
+	msgs, err := ch.Consume(
+		q.Name, // queue
+		"",     // consumer
+		true,   // auto-ack
```
subomi, line 114: @nitzangoldfeder, preferably, we should ack only when we've successfully written it to our internal queue. We should do this in the for loop when reading the message.

subomi (changes requested):
@nitzangoldfeder, FYI

internal/pkg/pubsub/amqp/client.go
```diff
@@ -0,0 +1,151 @@
...
+		if err := k.handler(ctx, k.source, string(d.Body)); err != nil {
+			k.log.WithError(err).Error("failed to write message to create event queue - amqp pub sub")
+			d.Ack(false)
+		} else {
+			// Reject the mesage and send it to DLQ
+			d.Nack(false, false)
+		}
+	}
+
+	<-forever
```
Dotunj, line 146: This will block forever till the process exists. I don't think we need this, since the `Consume()` function already runs in a goroutine. 

The cancellation context is also not being used to signal to the `Consume()` function when to return/end so that the process can release the `connection` and `channel` resources. 

One way to approach this is to check if `k.ctx` has been cancelled. 

```go
for d := range msgs {
	for d := range msgs {
		if errors.Is(k.ctx.Err(), context.Canceled) {
			return
		}

               // rest of code
	}
}
```

And then close the `connection` and `channel` before we return from the `Consume()` function.

```go
func (k *Amqp) Consume() {
	conn, err := k.dialer()
	if err != nil {
		log.WithError(err).Error("failed to instanciate a connection")
		return
	}


	ch, err := conn.Channel()
	if err != nil {
		log.WithError(err).Error("failed to instanciate a channel")
		return
	}

        defer conn.Close()
        defer ch.Close()

        // rest of code
}
```

internal/pkg/pubsub/amqp/client.go
```diff
@@ -0,0 +1,151 @@
...
+	)
+
+	if err != nil {
+		log.WithError(err).Error("failed to consume messages")
+		return
+	}
+
+	for d := range msgs {
+		ctx := context.Background()
+		if err := k.handler(ctx, k.source, string(d.Body)); err != nil {
```
subomi, line 138: @nitzangoldfeder, we should prefer `k.ctx` here instead of `context.Background()`

internal/pkg/pubsub/amqp/client.go
```diff
@@ -0,0 +1,151 @@
...
+		return err
+	}
+
+	ch, err := conn.Channel()
+	if err != nil {
+		log.WithError(err).Error("failed to instanciate a channel")
+		return err
+	}
+	defer ch.Close()
+
```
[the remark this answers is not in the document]
subomi: @nitzangoldfeder We closed `ch`, but we didn't close `conn` here.

subomi (changes requested):

subomi:
@Dotunj, this lgtm!, do you have any other requests?

Dotunj:
> @Dotunj, this lgtm!, do you have any other requests?

LGTM

subomi (approved):

Issue #1931, opened 2024-02-14
[Feature request]: Event rejection for unmatched filters


codeasashu:
### Is your feature request related to a problem? Please describe.
When a subscription doesn't match the event due to filters, ingest response still shows the event as accepted. This is frustrating. The panel doesn't show any event being accepted  (which is correct behaviour), and the ingest API also doesn't return any error (which is frustrating behaviour)

### Describe the solution you'd like
The ingest API should return an ERROR whenever event is discarded due to filter matching algorithm. I am guessing the filters are not matched straighaway, and the API may not have a way to know if it will match. One way to solve this is to return an identifier (similar to task-id in celery) and use it to know the event status. The workers can also update this identifier with status along the way.

### Describe alternatives you've considered
Don't know if there is any alternative to this

### Additional context
In celery, it is quite easy to track down a task. The "ingest" api in celery (which invokes the task) usually returns a task-id. The task id can then be used to further check the status of task. This is quite handy.

codeasashu:
@jirevwe I also see that a ingest event may have multiple deliveries. If the first queued delivery isn't success, it will be retired based on the setting. Hence, for an ingested event, we will have an array of deliveries, sorted by timestamps. If there are no forced retries, the latest delivery may have a success status.

In case of forced retries, there may be multiple "success" deliveries". In that case, the user may want to know if it was "forced", hence we can label them.

In conclusion, we can have an API which:
- list all the event deliveries
- able to filter delivery by status (succeeded, failed)
- filter by delivery sources ("forced" is kind of a source of delivery, ingest api is another)

What do you think?
I don't know golang much but I will give this a shot

PR #1933, opened 2024-02-17
add: eventId in ingest response

codeasashu:
This provides a parital fix to #1931 

the ingest response now contains `eventId` param, which can be used to:
1. Inspect the workers to see what has happened with event 
2. use in event api to see the past dispatches

Additional logging are added to filters, to see when an event didn't matched body/header filters.
Logging also have `event-id` keyword in JSON dict, to filter the logs out

### DIFF
--- a/api/ingest.go
+++ b/api/ingest.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"errors"
+
 	"github.com/frain-dev/convoy/pkg/msgpack"
 
 	"io"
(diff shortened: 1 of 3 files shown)


jirevwe:
Hey @codeasashu,

Did this change work for your usecase? We're do some re-architecting and would be moving code around, I'll keep this open and add the changes when I get around to working on it.

codeasashu:
Yes, it's not complete but it does work and seems like the way forward. I'd like to rebase at later point once I am done, if thats okay with you?

jirevwe:
Sure, that's fine. Thanks

PR #1936, opened 2024-02-20
Postgres connection string encoded credentials

matt-glover:
# Overview

This PR reworks the Postgres connection string builder to percent-encode the password portion of the string. I discovered while spinning up an environment that the connection string does not handle special characters in the password. I spotted this while using auto-generated passwords which I pull directly from a secrets manager into the convoy environment variables.

Previously convoy placed the raw password into a postgres URI: `postgres://username:raw_password@hostname:port/dbname?any_params`. If the password has any special characters, like `@` or `:` for example, it would fail further downstream in the app code / library code when it tried to process the postgres URI and found URI delimiters in the password. The errors seen vary depending on which special characters are used. The library code might think the hostname and/or port section starts early for example.

Per https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING we can expect these URIs to largely follow RFC 3986 which tells us to percent encode special characters. Percent encoding should eliminate this type of connection string construction error for passwords.

It's possible there are other parts of the URI that could run into issues with special characters but I kept this PR scoped to the password since that's the only problematic piece in my environment.

_Note: Anyone who manually percent-encoded a postgres password to get things running previously would consider this a breaking change._

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


subomi (approved):
Thank you @matt-glover

PR #1944, opened 2024-03-01
Fix Posthog Telemetry

jirevwe:


subomi:
We want to
- Make it hourly. 
- Add version number. 
- Use retained events.

subomi (approved):

PR #1962, opened 2024-03-30
fix: fixed subscriptions filtering in worker

subomi:


worker/task/process_event_creation.go
```diff
@@ -290,7 +290,7 @@ func findSubscriptions(ctx context.Context, endpointRepo datastore.EndpointRepos
 	return subscriptions, nil
 }
 
-func matchSubscriptionsUsingFilter(ctx context.Context, e *datastore.Event, subRepo datastore.SubscriptionRepository, subscriptions []datastore.Subscription) ([]datastore.Subscription, error) {
+func matchSubscriptionsUsingFilter(ctx context.Context, e *datastore.Event, subRepo datastore.SubscriptionRepository, subscriptions []datastore.Subscription, soft bool) ([]datastore.Subscription, error) {
```
jirevwe, line 293: What does soft do?

subomi: `soft` is asking the routine not to fail entirely because one subscription match failed. It's causing broadcast events to fail because one subscription match is failing.

jirevwe: Hmm, so if there are 10 subs and one fails, the whole thing fails?

subomi: Exactly.

jirevwe: Ah I see

jirevwe (approved):

PR #1992, opened 2024-05-03
bg: fix panic when request body is a string

olamilekan000:
```
change adds support for string request body in the flatten method.
```
Screenshot 2024-05-03 at 16 09 14


olamilekan000:
> Can you add some test cases for this?

done

jirevwe (approved):

PR #1995, opened 2024-05-06
fix: close res body

testwill:


subomi:
Thanks, @testwill, but that component is being deprecated.

PR #2002, opened 2024-05-14
Update rate limit function

jirevwe:
This PR aims to reduce the CPU usage of the `convoy.take_token()` function.

### DIFF
--- a/sql/1715118159.sql
+++ b/sql/1715118159.sql
@@ -1,0 +1,108 @@
+-- +migrate Up
+-- +migrate StatementBegin
+create or replace function convoy.take_token(_key text, _rate integer, _bucket_size integer) returns boolean
+    language plpgsql
+as
+$$
+DECLARE
+    next_min timestamptz;
+    _can_take BOOLEAN;
+    row RECORD;
+BEGIN
+    SELECT * FROM convoy.token_bucket WHERE key = _key FOR UPDATE SKIP LOCKED LIMIT 1 INTO row;
+    next_min := current_timestamp + make_interval(secs := _bucket_size);
+
+    IF current_timestamp < row.expires_at AND row.tokens = _rate THEN
+        RETURN FALSE;
+    END IF;
+
+    -- Update existing record
+    UPDATE convoy.token_bucket
+    SET tokens =
+            CASE WHEN current_timestamp > expires_at
+                     THEN 1
+                 ELSE CASE WHEN tokens < _rate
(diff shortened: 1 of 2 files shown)


sql/1715118159.sql
```diff
@@ -0,0 +1,108 @@
...
+create or replace function convoy.take_token(_key text, _rate integer, _bucket_size integer) returns boolean
+    language plpgsql
+as
+$$
+DECLARE
+    next_min timestamptz;
+    _can_take BOOLEAN;
+    row RECORD;
+BEGIN
+    SELECT * FROM convoy.token_bucket WHERE key = _key FOR UPDATE SKIP LOCKED LIMIT 1 INTO row;
```
subomi, line 12: @jirevwe nitpick, but for perf reasons, we should only select the columns we need. :)

sql/1715118159.sql
```diff
@@ -0,0 +1,112 @@
...
+                           ELSE expires_at
+                     END
+                END,
+        rate = COALESCE(_rate, rate),
+        updated_at = DEFAULT
+    WHERE key = _key
+    RETURNING TRUE INTO can_take;
+
+    -- Insert if no record found
+    IF NOT FOUND THEN
```
subomi, line 46: @jirevwe, what if we bootstrap the row for each endpoint at app startup? We won't need this anymore.

jirevwe: The table is unlogged. It will be wiped if the VM it's running on is restarted or recycled. So it's best to create the record lazily.

sql/1715118159.sql
```diff
@@ -0,0 +1,112 @@
...
+    language plpgsql
+as
+$$
+DECLARE
+    next_min timestamptz;
+    can_take BOOLEAN;
+    row RECORD;
+BEGIN
+    SELECT expires_at, tokens FROM convoy.token_bucket WHERE key = _key FOR UPDATE SKIP LOCKED LIMIT 1 INTO row;
+    if row is null then
```
subomi, line 13: @jirevwe, I'm probably missing something here, but we're unable to differentiate between if the row is locked by another transaction or if the row does not exist at all, so the `IF NOT FOUND` below will never run.

jirevwe: `if row is null then` is for the select query.

`IF NOT FOUND THEN` captures the updated row from the update query. If there was no update it'll run the insert query.

PR #2005, opened 2024-05-15
added metrics for the data plane

mekilis:


database/postgres/postgres_collector.go
```diff
@@ -0,0 +1,294 @@
...
+	)
+
+	eventQueueBacklogDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, "", "event_queue_backlog_seconds"),
+		"Number of seconds the oldest pending task is waiting in pending state to be processed.",
+		[]string{"project_id", "source"}, nil,
+	)
+
+	eventDeliveryQueueTotalDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, "", "convoy_event_delivery_queue_total"),
```
subomi, line 82: @mekilis, this shouldn't have the prefix `convoy` since it's already in the namespace.

database/postgres/postgres_collector.go
```diff
@@ -0,0 +1,294 @@
...
+	"github.com/frain-dev/convoy/pkg/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"strings"
+	"time"
+)
+
+// Namespace used in fully-qualified metrics names.
+const namespace = "convoy"
+
+const delaySeconds = 1
```
subomi, line 13: @mekilis, the sample time should be exposed as an instance-level configuration.

subomi: @mekilis, because this can have a significant performance impact, we should add an instance-level configuration to enable/disable metrics. Something like:

```json
{
  "metrics": {
    "backend": "prometheus",
    "prometheus": {
      // sample time in seconds.
      "sample_time": 10 
    }
  }
}
```

mekilis: Got it.

mekilis: Done

database/postgres/postgres_collector.go
mekilis: Done

subomi (approved):

PR #2011, opened 2024-05-17
Move retention policy to instance config

danvixent:


api/migrations/v20240517/projects.go
```diff
@@ -0,0 +1,87 @@
+package v20240517
```
subomi, line 1: @danvixent, we don't need to make this backwards compatible.

danvixent: @subomi okay!

cmd/hooks/hooks.go
```diff
@@ -175,14 +177,16 @@ func PreRun(app *cli.App, db *postgres.Postgres) func(cmd *cobra.Command, args [
...
 		// update config singleton with the instance id
-		configRepo := postgres.NewConfigRepo(app.DB)
-		instCfg, err := configRepo.LoadConfiguration(cmd.Context())
-		if err != nil {
-			log.WithError(err).Error("Failed to load configuration")
-		} else {
-			cfg.InstanceId = instCfg.UID
-			if err = config.Override(&cfg); err != nil {
-				return err
+		if cmd.Use != "bootstrap" && cmd.Use != "version" { // bootstrap & version don't need this
```
jirevwe, line 180: Hmm, nice. I think `migrate` doesn't need this too. Can you add it?

config/config.go
```diff
@@ -251,6 +255,11 @@ type SentryConfiguration struct {
 	DSN string `json:"dsn" envconfig:"CONVOY_SENTRY_DSN"`
 }
 
+type RetentionPolicyConfiguration struct {
```
jirevwe, line 258: Didn't we have this before? Or was it only in the project settings?

internal/pkg/exporter/exporter.go
```diff
@@ -232,9 +234,27 @@ func getOutputWriter(out string) (io.WriteCloser, error) {
...
+//}
+//
+//func (c compressWriter) Close() error {
+//	if err := c.gw.Close(); err != nil {
+//		return err
+//	}
+//
+//	return c.file.Close()
+//}
+
```
jirevwe, line 257: Shouldn't we remove this for now?

danvixent: I figure we will need it in the future

config/config.go
danvixent: Only project settings

jirevwe: ok

cmd/hooks/hooks.go
danvixent: Done

PR #2033, opened 2024-05-29
Add indexes for FetchSubscriptionsForBroadcast hot path

danvixent:


jirevwe:
The migration file with the indexes is missing

jirevwe (approved):

PR #2049, opened 2024-06-01
feat: re-add in-memory subscriptions loader

subomi:


database/postgres/subscription.go
```diff
@@ -121,6 +122,40 @@ const (
...
+
+	fetchDeletedSubscriptions = `
+    select name, id, type, project_id, endpoint_id, function, updated_at,
+    filter_config_event_types AS "filter_config.event_types",
+    filter_config_filter_headers AS "filter_config.filter.headers",
+	filter_config_filter_body AS "filter_config.filter.body"
+    from convoy.subscriptions
+    where (deleted_at > $4 AND deleted_at is not null)
+    AND id > $1
+    AND project_id = $2
```
danvixent, line 158: @subomi please make a base fetch constant query, that way we can avoid any one of them just randomly changing.

internal/pkg/memorystore/table.go
```diff
@@ -85,18 +85,34 @@ func (t *Table) Add(key string, value interface{}) *Row {
...
 
+// Add a new item if it doesn't exist.
+func (t *Table) Upsert(key string, value interface{}) *Row {
+	t.Lock()
+	defer t.Unlock()
+
+	row := t.getInternal(key)
+	if row != nil {
+		t.deleteInternal(key)
+	}
```
danvixent, line 119: The row key will be updated or created by the t.rows map, extra check not needed.

database/postgres/subscription.go
subomi: @danvixent I'll defer this for a later refactor.

worker/task/process_broadcast_event_creation.go
```diff
@@ -76,6 +76,7 @@ func ProcessBroadcastEventCreation(endpointRepo datastore.EndpointRepository, ev
 		es, ss := getEndpointIDs(subscriptions)
 		event.Endpoints = es
 
+		fmt.Println(es)
```
jirevwe, line 79: Remove print

danvixent (approved):

PR #2064, opened 2024-06-07
changed postgres driver to pgxpool

mekilis:


database/postgres/postgres.go
```diff
@@ -15,7 +16,6 @@ import (
 	"github.com/frain-dev/convoy/pkg/log"
 	"github.com/jmoiron/sqlx"
 	_ "github.com/newrelic/go-agent/v3/integrations/nrpq"
```
jirevwe, line 18: I don't think we need this import anymore either

mekilis: True

.gitignore
```diff
@@ -35,6 +35,6 @@ convoy.json
 docker-compose.yml
 compose.yml
 typesense_data
-postgres_data
+postgres_data**
```
jirevwe, line 38: Won't the below work instead?
```suggestion
postgres_data/
```

jirevwe:
It looks good; we need to test it to see if it performs better.

mekilis:
> It looks good; we need to test it to see if it performs better.

Yea. Still need to fix failing tests

api/server_suite_test.go
```diff
@@ -62,17 +63,23 @@ func getConfig() config.Configuration {
...
+		}
+		_ = os.Setenv("TZ", "") // Use UTC by default :)
 
-	dbHooks := hooks.Init()
-	dbHooks.RegisterHook(datastore.EndpointCreated, func(data interface{}, changelog interface{}) {})
+		dbHooks := hooks.Init()
+		dbHooks.RegisterHook(datastore.EndpointCreated, func(data interface{}, changelog interface{}) {})
 
-	return db
+		pDB = db
```
danvixent, line 80: Hmm i'm a bit confused by this func

mekilis: The tests were running out of db connections.

jirevwe (approved):

PR #2079, opened 2024-06-28
added end to end latency metrics

mekilis:


worker/task/process_event_creation.go
```diff
@@ -221,6 +221,7 @@ func writeEventDeliveriesToQueue(ctx context.Context, subscriptions []datastore.
 			payload := EventDelivery{
 				EventDeliveryID: eventDelivery.UID,
 				ProjectID:       eventDelivery.ProjectID,
+				AcknowledgedAt:  eventDelivery.AcknowledgedAt,
```
jirevwe, line 224: Why are we passing this in the queue payload?

mekilis: The payload is an `EventDelivery` object which is a minified version of the actual eventDelivery, so it should still have the `acknowledgedAt` field in case it is needed at the other end.

jirevwe: I don't think we need it till we dispatch the event delivery. What if we fetch the ack time from the event delivery's event?

mekilis: Ok. Will remove.

jirevwe (approved):

database/postgres/event_delivery.go
```diff
@@ -979,8 +983,10 @@ type EventDeliveryPaginated struct {
 	Headers        httpheader.HTTPHeader `json:"headers" db:"headers"`
 	URLQueryParams string                `json:"url_query_params" db:"url_query_params"`
 	IdempotencyKey string                `json:"idempotency_key" db:"idempotency_key"`
-	Latency        string                `json:"latency" db:"latency"`
-	EventType      datastore.EventType   `json:"event_type,omitempty" db:"event_type"`
+	// Deprecated: Latency is deprecated.
+	Latency        string              `json:"latency" db:"latency_seconds"`
```
mekilis, line 987: Why is the `db` not `latency`?

config/config.go
```diff
@@ -387,7 +387,7 @@ type Configuration struct {
 	ConsumerPoolSize    int                        `json:"consumer_pool_size" envconfig:"CONVOY_CONSUMER_POOL_SIZE"`
 	EnableProfiling     bool                       `json:"enable_profiling" envconfig:"CONVOY_ENABLE_PROFILING"`
 	Metrics             MetricsConfiguration       `json:"metrics" envconfig:"CONVOY_METRICS"`
-	InstanceIngestRate  int                        `json:"pub_sub_ingest_rate" envconfig:"CONVOY_PUB_SUB_INGEST_RATE"`
+	InstanceIngestRate  int                        `json:"instance_ingest_rate" envconfig:"CONVOY_INSTANCE_INGEST_RATE"`
```
danvixent, line 390: This will break things in cloud

database/postgres/event.go
```diff
@@ -37,7 +37,8 @@ const (
     raw, data, headers, is_duplicate_event,
 	COALESCE(source_id, '') AS source_id,
 	COALESCE(idempotency_key, '') AS idempotency_key,
-	COALESCE(url_query_params, '') AS url_query_params
+	COALESCE(url_query_params, '') AS url_query_params,
+	created_at,updated_at,acknowledged_at
```
danvixent, line 41: Why are we adding created_at,updated_at columns to events_endpoints table?

database/postgres/subscription.go
```diff
@@ -288,6 +288,10 @@ func (s *subscriptionRepo) FetchDeletedSubscriptions(ctx context.Context, projec
 }
 
 func (s *subscriptionRepo) LoadAllSubscriptionConfig(ctx context.Context, projectIDs []string, pageSize int64) ([]datastore.Subscription, error) {
+	if len(projectIDs) == 0 {
+		projectIDs = []string{""}
+	}
+
```
danvixent, line 294: projectIDs is a compulsory filter, if it's empty it should return empty subscriptions.

database/postgres/subscription.go
```diff
@@ -377,6 +381,10 @@ func (s *subscriptionRepo) FetchSubscriptionsForBroadcast(ctx context.Context, p
 }
 
 func (s *subscriptionRepo) fetchChangedSubscriptionConfig(ctx context.Context, countQuery, query string, projectIDs []string, t time.Time, pageSize int64) ([]datastore.Subscription, error) {
+	if len(projectIDs) == 0 {
+		projectIDs = []string{""}
+	}
+
```
danvixent, line 387: Same here i think

database/postgres/event.go
```diff
@@ -24,8 +24,8 @@ const (
 	createEvent = `
 	INSERT INTO convoy.events (id,event_type,endpoints,project_id,
 	                           source_id,headers,raw,data,url_query_params,
-	                           idempotency_key,is_duplicate_event,created_at,updated_at)
-	VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13)
+	                           idempotency_key,is_duplicate_event,acknowledged_at)
```
danvixent, line 27: I'm a bit concerned about this, i agree it's better to use `acknowledged_at` at the entry point when we queue the event helps to record latency correctly, but as at the point when we insert it into the db, isn't now in the true sense, created? By this logic I think `created_at,updated_at` should still be inserted here.

mekilis: The db should tell when it is created. At this point, it is source of truth.

database/postgres/event.go
mekilis: No, this is `fetchEventById`, unrelated to endpoints.

database/postgres/subscription.go
jirevwe: But that's not how the query works. When the db is empty (no projects yet) the service doesn't start up

config/config.go
jirevwe: We can change it before it gets to the cloud; I don't see it being an issue.

jirevwe: I made the change in my rate-limiter PR, but some other PR reverted it that's why I made it here again.

jirevwe: It's mainly because of [this issue](https://github.com/jmoiron/sqlx/issues/196)

database/postgres/subscription.go
jirevwe: It's because of [this issue](https://github.com/jmoiron/sqlx/issues/196); we use `IN (?)` in the query

danvixent: @jirevwe that's my point exactly, this empty filter now won't match anything in db, we should just do this:

```go
if len(projects) == 0 {
    return []datastore.Subscription{}, nil
}
```

config/config.go
danvixent: Okay

database/postgres/event.go
danvixent: @mekilis i don't understand, are we not saying the same thing?

database/postgres/subscription.go
jirevwe: True true

database/postgres/subscription.go
jirevwe: Hmm ok

danvixent (approved):

PR #2088, opened 2024-07-11
Update event data openapi types

balzdur:
# Description

Generate better OpenAPI API models for `json.RawMessage` and change some `200` to the correct `201` status code.

# Motivation

**TL;DR:** By using the generated openapi spec to generate a go api client, I found a weird model generation for event data. Reading previous changes done in [this PR](https://github.com/frain-dev/convoy/pull/1959) on the `models.CreateEvent` struct make me push those changes.

In order to use `convoy` in a Go project, we tried to make use of the `confoy-go` SDK. Unfortunately, after some experimentation we found the SDK is not up to date with the current API (ex: RetryConfig in Responses is not up to data, leading in parsing errors in runtime). Since the OpenAPI spec is generated on each api changes, we want to generate a Convoy API client using [this project](https://github.com/oapi-codegen/oapi-codegen). Unfortunately, some event data type are inferred as `Data *[]int` based on the OAS spec. Those changes fix my issue (generated as `Data *map[string]interface{}`)

# Proposed solution

By digging `oapi-codegen` issues I found using the `swaggertype:"object"` should do the trick. This change seems to have already been applied on `models.CreateEvent` in [this PR](https://github.com/frain-dev/convoy/pull/1959), that's why I propose to generalise it to the whole file.

> NB: part of the doc changes are due to previous changes  on `main` not present in the doc yet


danvixent:
@balzdur Thanks for reporting this!

danvixent (approved):

PR #2091, opened 2024-07-15
docs: update api docs success response

balzdur:
Same as previous one, I found some errors in success return codes


mekilis (approved):
Thanks!

PR #2095, opened 2024-07-18
fix: Convoy should respond with a 413 for oversized payloads

### LINKED ISSUE #2094: [Bug report]: Convoy should respond with a 413 for oversized payloads
### Describe the bug
There is a Convoy configuration setting `max_response_size` that limits the size of client requests sent to ingest, but when a client sends a payload that is too large, Convoy still responds with a `200`.

### To Reproduce
1. Generate a sample payload larger than the configured `max_response_size`
2. `POST` it to the Convoy subscription endpoint
3. Convoy will respond with `{"status":true,"message":"Event received","data":51200}` and an http code of `200`.

### Expected behavior
Convoy would ideally respond with a 413 (Payload Too Large)

### Stack (please complete the following information):
 - Convoy: v24.1.2
 - PostgreSQL: 15
 - Redis: 7
 - OS: Kubernetes

BolajiOlajide:
Closes #2094

This ensures the `ingest` endpoint always respect the max size accepted for payloads.


scripts/integration-test.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
BolajiOlajide, line 1: Added this to make editors understand that this is a shell script. Also made it an executable, for some reason it wasn't.

jirevwe (approved):

danvixent (approved):

PR #2105, opened 2024-07-22
refactored and feature flags per feature

mekilis:


cmd/main.go
```diff
@@ -96,7 +98,8 @@ func main() {
 	c.Flags().StringVar(&redisDatabase, "redis-database", "", "Redis database")
 	c.Flags().IntVar(&redisPort, "redis-port", 0, "Redis Port")
 
-	c.Flags().StringVar(&fflag, "feature-flag", "", "Enable feature flags (experimental)")
+	c.Flags().StringSliceVar(&fflag, "enable-feature-flag", []string{}, "List of feature flags to enable e.g. \"search_tokenizer,prometheus\"")
```
subomi, line 101: @mekilis Can we change this from `search_tokenizer` to `webhook-full-text-search`.

cmd/main.go
```diff
@@ -96,7 +98,8 @@ func main() {
...
+	c.Flags().StringSliceVar(&noFflag, "disable-feature-flag", []string{}, "List of feature flags to disable e.g. \"search_tokenizer,prometheus\"")
```
subomi, line 102: @mekilis We don't need a `disable-feature-flag` flag. Experimental features should be disabled by default.

jirevwe (approved):

cmd/main.go
subomi: Ok. Should we use `-` or `_` as the delimiter? I prefer `-`.

cmd/main.go
subomi: @mekilis Yeah, but most folks don't use it yet, so when they upgrade they'd need to explicitly enable it otherwise it shouldn't work.

PR #2153, opened 2024-09-20
allow only portal link load in iframe

Oluwadaminiola:


web/ui/dashboard/src/app/guards/iframe/iframe.guard.ts
```diff
@@ -0,0 +1,13 @@
+import { Injectable } from '@angular/core';
+import { CanActivate } from '@angular/router';
+
+@Injectable({
+	providedIn: 'root'
+})
+export class IframeGuard implements CanActivate {
+	canActivate(): boolean {
+		const isIframe = window.self !== window.top;
+		if (isIframe) return false;
```
jirevwe, line 10: Seems like this can be simplified to 
```suggestion
		return window.self === window.top;
```

jirevwe (approved):

PR #2166, opened 2024-10-04
Add HTTP API Rate Limit

jirevwe:


config/config.go
```diff
@@ -363,6 +364,7 @@ type Configuration struct {
 	EnableProfiling     bool                         `json:"enable_profiling" envconfig:"CONVOY_ENABLE_PROFILING"`
 	Metrics             MetricsConfiguration         `json:"metrics" envconfig:"CONVOY_METRICS"`
 	InstanceIngestRate  int                          `json:"instance_ingest_rate" envconfig:"CONVOY_INSTANCE_INGEST_RATE"`
+	HttpApiRateLimit    int                          `json:"http_api_rate_limit" envconfig:"CONVOY_HTTP_API_RATE_LIMIT"`
```
subomi, line 367: @jirevwe Do we need the `http_` prefix here? I think it's okay to remove it.

subomi:
@jirevwe There's no corresponding command line config, either.

config/config.go
jirevwe: Done

subomi (approved):

PR #2169, opened 2024-10-14
Add allow and block IP lists

jirevwe:
This PR introduces a way to configure IP allow and block lists for outgoing webhooks without using a separate software (egress proxy, e.g., Smokescreen).

From a layered security perspective, this implementation:
- offers greater security by baking in IP filtering rules.
- allows granular control over network requests.
- allows flexible configuration for specific webhook gateway components.

### Usage
It is currently put behind a feature flag, so usage will require the flag to be explicitly specified when running Convoy.

```shell
convoy agent --enable-feature-flag "ip-rules" --ip-allow-list "3.14.163.112/32,3.217.134.199/32"
```

It is also behind a business license. Hence, a license key must be explicitly specified when running Convoy.

```shell
convoy agent --enable-feature-flag --license-key "ABC-XYZ-123" "ip-rules" --ip-block-list "3.14.163.112/32,3.217.134.199/32"
```


config/config.go
```diff
@@ -388,6 +393,13 @@ type Configuration struct {
 	WorkerExecutionMode ExecutionMode                `json:"worker_execution_mode" envconfig:"CONVOY_WORKER_EXECUTION_MODE"`
 	MaxRetrySeconds     uint64                       `json:"max_retry_seconds,omitempty" envconfig:"CONVOY_MAX_RETRY_SECONDS"`
 	LicenseKey          string                       `json:"license_key" envconfig:"CONVOY_LICENSE_KEY"`
+	Dispatcher          DispatcherConfiguration      `json:"dispatcher"`
+}
+
+type DispatcherConfiguration struct {
+	EnforceSecure bool     `json:"enforce_secure" envconfig:"CONVOY_DISPATCHER_ENFORCE_SECURE"`
```
subomi, line 400: @jirevwe We have a project-level configuration for this. I don't think we need it here as well.

config/config.go
```diff
@@ -388,6 +393,13 @@ type Configuration struct {
...
+	AllowList     []string `json:"allow_list" envconfig:"CONVOY_DISPATCHER_ALLOW_LIST"`
+	BlockList     []string `json:"block_list" envconfig:"CONVOY_DISPATCHER_BLOCKLIST"`
```
subomi, line 402: @jirevwe Nitpick, but we should try to be consistent here: `ALLOW_LIST` and `BLOCK_LIST`.

net/dispatcher.go
```diff
@@ -19,44 +23,158 @@ import (
...
 	}
+}
 
-	if licenser.UseForwardProxy() {
-		proxyUrl, isValid, err := d.setProxy(httpProxy)
-		if err != nil {
+// AllowListOption sets a list of IP prefixes which will outgoing traffic will be granted access
+func AllowListOption(allowList []string) DispatcherOption {
+	return func(d *Dispatcher) error {
+		if len(allowList) == 0 {
```
subomi, line 67: @jirevwe Is there a reason these values can't be empty? Making it required can be a proactive measure to keep our users safe. But this will break a lot of prod deployments.

net/dispatcher.go
```diff
@@ -19,44 +23,158 @@ import (
...
+)
+
+type DispatcherOption func(d *Dispatcher) error
+
 type Dispatcher struct {
-	client *http.Client
+	logger    *log.Logger
+	transport *http.Transport
+	client    *http.Client
+	rules     *netjail.Rules
```
subomi, line 39: @jirevwe Also, `rules` should be behind the license gate.

net/dispatcher.go
```diff
@@ -19,44 +23,158 @@ import (
...
+
+	for _, option := range options {
+		if err := option(d); err != nil {
 			return nil, err
 		}
+	}
 
-		if isValid {
-			tr.Proxy = http.ProxyURL(proxyUrl)
+	if len(d.rules.Allow) == 0 && len(d.rules.Block) == 0 {
```
subomi, line 154: @jirevwe What happens if only one is set? From what I'm seeing, it will eventually fail, which is not what we want. I don't think either of them should be required. It shouldn't work differently from how it works in production today. What do you think?

config/config.go
jirevwe: The dispatcher is created once and reused multiple times because the client and its transport need to be "created once and reused multiple times" so we can take advantage of connection reuse. If we configure it per project, we'd need to toggle it on and off at runtime, which might be messy.

config/config.go
jirevwe: Seen

net/dispatcher.go
jirevwe: Got it

config/config.go
subomi: But we can't have both. We either remove it from the project settings or here. Another suggestion is to reuse the client and transport on multiple dispatcher instances.

danvixent (approved):

mekilis (approved):
Fine!

PR #2175, opened 2024-10-23
added login with SSO

mekilis:


api/handlers/auth.go
```diff
@@ -15,6 +15,71 @@ import (
 	"github.com/go-chi/render"
 )
 
+func (h *Handler) InitLoginSSO(w http.ResponseWriter, r *http.Request) {
+
+	configuration, err := config.Get()
+	if err != nil {
+		_ = render.Render(w, r, util.NewErrorResponse(err.Error(), http.StatusBadRequest))
+		return
+	}
```
jirevwe, line 24: The handler already has a copy of the configuration. It should be h.A.Cfg

mekilis: Noted, thank you.

services/login_sso.go
```diff
@@ -0,0 +1,236 @@
...
+	if !tokenResponse.Status {
+		log.Errorf("failed to redeem token in response: %+v", tokenResponse)
+		return nil, errors.New("failed to redeem token")
+	}
+
+	if tokenResponse.Data.Payload.Email == "" {
+		return nil, errors.New("no email in token response")
+	}
+	if tokenResponse.Data.Payload.OrganizationExternalID == "" {
+		return nil, errors.New("no email in token response")
```
danvixent, line 144: I'm not sure if this error message is on purpose?

mekilis: It's not. An oversight that will be corrected

jirevwe (approved):

PR #2190, opened 2024-11-13
added datadog tracer

mekilis:


internal/pkg/tracer/tracer.go
```diff
@@ -33,33 +41,55 @@ func FromContext(ctx context.Context) trace.Tracer {
...
-			return noopShutdownFn, ErrInvalidTracerConfiguration
+			return st, ErrInvalidTracerConfiguration
+		}
+
+		return st, st.Init(componentName)
+
+	case config.DatadogTracerProvider:
+		dt := NewDatadogTracer(tCfg.Datadog, licenser)
+		if !licenser.DatadogTracing() {
+			log.Error(ErrTracerFeatureUnavailable.Error())
```
jirevwe, line 76: We should pass the log instance so this line respect the configured log level

jirevwe (approved):

internal/pkg/tracer/tracer.go
mekilis: Noted

PR #2202, opened 2024-12-07
added credential encryption lifecycle management

mekilis:
This PR implements a comprehensive credential encryption lifecycle for the `endpoints` table (which can be extended later to other tables), focusing on enhanced security for sensitive data while providing flexibility through initialization, rotation, and reversion workflows. The feature is gated behind a licensing check to ensure controlled usage. 

Key updates include:

#### Encryption Initialization:

Introduced a process to initialize encryption for existing plaintext credentials in the endpoints table. New `_cipher` columns (`secrets_cipher` and `authentication_type_api_key_header_value_cipher`) store encrypted versions of the `secrets` and `authentication_type_api_key_header_value` columns.
Once encryption is applied, original plaintext columns are reset to zero values (`NULL` or `'[]'` for JSONB, depending on constraints).

```bash
convoy utils init-encryption <encryption-key> --enable-feature-flag="credential-encryption"
```

#### Key Rotation:

Added a mechanism to rotate encryption keys seamlessly. The system decrypts existing encrypted data, re-encrypts it using the new key, and updates the `_cipher` columns accordingly.

```bash
convoy utils rotate-key <old-key> <new-key> --enable-feature-flag="credential-encryption"
```

#### Reversion to Plaintext:

Implemented a reversion feature to decrypt and restore credentials into their plaintext columns if encryption is disabled or the license is revoked. Ensures smooth transitions without data loss or corruption.

```bash
convoy utils revert-encryption <encryption-key> --enable-feature-flag="credential-encryption"
```

#### Query Updates:

Enhanced `SELECT`, `INSERT`, and `UPDATE` queries to dynamically handle encryption and decryption based on the `is_encrypted` flag. Ensures backward compatibility for non-encrypted environments.

#### Feature Licensing:

Encryption workflows are gated behind a licensing check, ensuring the feature is only accessible to users with the required license. Systems without the license will continue to function using plaintext credentials.

#### Other Notes:
The default backend for storing the encryption keys is HCP Vault Secrets. 

##### Configuration (JSON): 
```json
    "hcp_vault": {
        "client_id": "CLIENT_ID",
        "client_secret": ".CLIENT_SECRET",
        "org_id": "ORG_ID",
        "project_id": "PROJECT_ID",
        "app_name": "sample-app",
        "secret_name": "static_secret",
        "cache_duration": 2
    },
```

The feature also requires the feature flag to be enabled:
```
--enable-feature-flag="credential-encryption"
```


cmd/main.go
```diff
@@ -135,6 +136,8 @@ func main() {
 
 	c.Flags().Uint64Var(&maxRetrySeconds, "max-retry-seconds", 7200, "Max retry seconds exponential backoff")
 
+	AddHCPVaultFlags(c)
```
jirevwe, line 139: Is there a need to include them in a function like this?

cmd/utils/revert_encryption.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/frain-dev/convoy/internal/pkg/keys"
+	"github.com/frain-dev/convoy/pkg/log"
+	"github.com/spf13/cobra"
+)
+
+func AddRevertEncryptionCommand(a *cli.App) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "revert-encryption <encryption-key>",
+		Short: "Reverts the encryption initialization for the specified table columns with the provided encryption key",
+		Args:  cobra.ExactArgs(1),
```
jirevwe, line 18: Do we need to include this? Since we have already retrieved the key from line 20. Or is that how it's meant to work?

cmd/utils/revert_encryption.go
```diff
@@ -0,0 +1,63 @@
...
+
+			if encryptionKey == "" {
+				return fmt.Errorf("encryption key cannot be empty")
+			}
+
+			log.Infof("Reverting encryption with the provided key...")
+
+			cfg, err := config.Get()
+			if err != nil {
+				log.WithError(err).Fatal("Error fetching the config.")
```
jirevwe, line 30: We should return an error here and not panic using Fatal

cmd/utils/revert_encryption.go
```diff
@@ -0,0 +1,63 @@
...
+			if encryptionKey != currentKey {
+				if !errors.Is(err, keys.ErrCredentialEncryptionFeatureUnavailable) {
+					return fmt.Errorf("provided encryption key does not match the current encryption key")
+				}
+				// allow any key if downgraded
+			}
+
+			db, err := postgres.NewDB(cfg)
+			if err != nil {
+				log.Fatal(err)
```
jirevwe, line 54: We should return an error here and not panic using Fatal

cmd/utils/revert_encryption.go
```diff
@@ -0,0 +1,63 @@
...
+			}
+			defer db.Close()
+
+			return keys.RevertEncryption(db, km, encryptionKey)
```
jirevwe, line 58: We should handle the error (potentially logging out that the revert failed) and then return a nil error value

jirevwe:
We should return error values and not exit the process using `Fatal()`

jirevwe:
The way we return errors is inconsistent. Some use `fmt.Errorf()` even when they don't do any formatting and some use `errors.New()`. The one's that return `errors.New()` can be extracted to variables for better readability, and the ones that use `fmt.ErrorF()` but don't do any formatting should be converted to `errors.New()`.

cmd/hooks/hooks.go
```diff
@@ -823,3 +828,64 @@ func closeWithError(closer io.Closer) {
 		fmt.Printf("%v, an error occurred while closing the client", err)
```
jirevwe, line 828: Can you help update this line to use the logger if it's in scope?

cmd/utils/rotate_key.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/frain-dev/convoy/internal/pkg/keys"
+	"github.com/frain-dev/convoy/pkg/log"
+	"github.com/spf13/cobra"
+)
+
+func AddRotateKeyCommand(a *cli.App) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "rotate-key <old-key> <new-key>",
+		Short: "Rotates the encryption key by re-encrypting data with a new key",
+		Args:  cobra.ExactArgs(2),
```
jirevwe, line 18: I see, so this is how it works; my previous comment might be invalid then.

cmd/utils/rotate_key.go
```diff
@@ -0,0 +1,73 @@
...
+
+			return keys.RotateEncryptionKey(db, km, oldKey, newKey)
+		},
+	}
+	return cmd
+}
+
+func validateAndGetKeys(args []string) (string, string, error) {
+	oldKey := args[0]
+	newKey := args[1]
```
jirevwe, line 64: if either of these are nil, will the program panic? Does cobra guarantee that the slice will contain nil values in args since we passed like this `cobra.ExactArgs(2)`?

config/config.go
```diff
@@ -423,6 +424,16 @@ type PyroscopeConfiguration struct {
...
 }
 
+type HCPVaultConfig struct {
+	ClientID      string        `json:"client_id" envconfig:"HCP_CLIENT_ID"`
+	ClientSecret  string        `json:"client_secret" envconfig:"HCP_CLIENT_SECRET"`
+	OrgID         string        `json:"org_id" envconfig:"HCP_ORG_ID"`
+	ProjectID     string        `json:"project_id" envconfig:"HCP_PROJECT_ID"`
+	AppName       string        `json:"app_name" envconfig:"HCP_APP_NAME"`
+	SecretName    string        `json:"secret_name" envconfig:"HCP_SECRET_NAME"`
+	CacheDuration time.Duration `json:"cache_duration" envconfig:"HCP_CACHE_DURATION"`
```
jirevwe, line 434: I think we need to prefix the env vars with `CONVOY_` to be consistent with the rest. I know we set a prefix but I'm not sure if it's automatically applied.

internal/pkg/keys/encrypter_init.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"fmt"
+	"github.com/frain-dev/convoy/database"
+	"github.com/frain-dev/convoy/pkg/log"
+	"github.com/jmoiron/sqlx"
+	"strings"
+)
+
+func InitEncryption(db database.Database, km KeyManager, encryptionKey string) error {
```
jirevwe, line 11: We should pass a log instance to this func so we can respect the currently running process' log level.

internal/pkg/keys/encrypter_init.go
```diff
@@ -0,0 +1,147 @@
...
+	tx, err := db.GetDB().Beginx()
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+
+	for table, columns := range tablesAndColumns {
+		log.Infof("Processing table: %s", table)
+
+		if err := lockTable(tx, table); err != nil {
+			_ = tx.Rollback()
```
jirevwe, line 22: We should handle the error from the transaction rollback or at least log it out. There's a function we created it's called `rollbackTx()` I think

internal/pkg/keys/encrypter_init.go
```diff
@@ -0,0 +1,147 @@
...
+
+// lockTable ensures the specified table is locked for exclusive access during the operation.
+func lockTable(tx *sqlx.Tx, table string) error {
+	// Set a statement timeout to avoid indefinite hanging on the lock
+	_, err := tx.Exec("SET statement_timeout = '120s';")
+	if err != nil {
+		return fmt.Errorf("failed to set statement timeout: %w", err)
+	}
+
+	lockQuery := fmt.Sprintf("LOCK TABLE convoy.%s IN ACCESS EXCLUSIVE MODE;", table)
```
jirevwe, line 86: Also, what if 2 mins isn't enough time?

jirevwe:
Some of the new queries added don't run with a context.

cmd/main.go
mekilis: Well, this portion was getting too long so I thought I should just modularize this config

cmd/utils/revert_encryption.go
mekilis: It's a sanity check. To be sure they know what they are doing.

cmd/utils/revert_encryption.go
mekilis: This should ideally never return an error

cmd/utils/revert_encryption.go
mekilis: Hmm. Ok

cmd/utils/revert_encryption.go
mekilis: Ok

cmd/hooks/hooks.go
mekilis: Ok

cmd/utils/rotate_key.go
mekilis: Which one though

cmd/utils/rotate_key.go
mekilis: Yes.

internal/pkg/keys/encrypter_init.go
mekilis: Ok

internal/pkg/keys/encrypter_init.go
mekilis: OK. Will check

internal/pkg/keys/encrypter_init.go
mekilis: It's a safe bet. It's even too long.

internal/pkg/keys/encrypter_init.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil && err.Error() != "sql: no rows in result set" {
+		return false, fmt.Errorf("failed to check encryption status of table %s: %w", table, err)
+	}
+	return isEncrypted, nil
+}
+
+// lockTable ensures the specified table is locked for exclusive access during the operation.
+func lockTable(tx *sqlx.Tx, table string) error {
+	// Set a statement timeout to avoid indefinite hanging on the lock
+	_, err := tx.Exec("SET statement_timeout = '120s';")
```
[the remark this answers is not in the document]
mekilis: Hmm. Context with timeout?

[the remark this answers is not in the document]
jirevwe: Nah, we just need to use the context being passed down so if we cancel it from the parent it will also cancel.

cmd/utils/revert_encryption.go
jirevwe: True, it's a nitpick, if anything.

cmd/utils/revert_encryption.go
jirevwe: Got it

cmd/main.go
jirevwe: Got it

jirevwe (approved):

PR #2206, opened 2024-12-20
Remove cache implementation

jirevwe:


database/postgres/event.go
```diff
@@ -160,18 +157,21 @@ const (
...
+		LEFT JOIN convoy.events_endpoints ee ON ev.id = ee.event_id
+		WHERE ev.deleted_at IS NULL
 	`
 
 	baseCountPrevEventSearch = `
-	SELECT COUNT(DISTINCT(ev.id)) AS COUNT
-	FROM convoy.events_search ev
-	LEFT JOIN convoy.events_endpoints ee ON ev.id = ee.event_id
-	WHERE ev.deleted_at IS NULL
+	select exists(
```
mekilis, line 168: They are equivalent, yeah?

services/create_endpoint.go
```diff
@@ -12,15 +12,14 @@ import (
...
-	"github.com/frain-dev/convoy/cache"
 	"github.com/frain-dev/convoy/datastore"
 	"github.com/frain-dev/convoy/pkg/log"
 	"github.com/frain-dev/convoy/util"
 	"github.com/oklog/ulid/v2"
 )
 
 type CreateEndpointService struct {
-	Cache          cache.Cache
+	// Cache          cache.Cache
```
mekilis, line 22: Should be deleted

mekilis (commented):
It's ok. Left two comments.

database/postgres/event.go
jirevwe: Yep

mekilis (approved):
Good to go.

PR #2232, opened 2025-02-03
Fix panic in backup project data handler

jirevwe:


internal/pkg/exporter/exporter.go
```diff
@@ -68,16 +68,11 @@ func NewExporter(projectRepo datastore.ProjectRepository,
...
-	if err != nil {
-		return nil, err
-	}
-
 	return &Exporter{
 		config:  c,
 		project: p,
 		result:  ExportResult{},
-		expDate: time.Now().UTC().Add(-policy),
+		expDate: time.Now().Add(-time.Hour * 24),
```
mekilis, line 75: Why this change?

mekilis (approved):
It's fine.

internal/pkg/exporter/exporter.go
jirevwe: The plan is to back up events sent today at the end of the day rather than at the end of the retention window because at the end of the policy window is when we would be deleting them.

PR #2244, opened 2025-02-19
Re-enable Meta-Events

jirevwe:


sql/1739961493.sql
```diff
@@ -0,0 +1,83 @@
...
+            ON DELETE CASCADE;
+
+ALTER TABLE convoy.sources
+    ADD CONSTRAINT sources_source_verifier_id_fkey
+        FOREIGN KEY (source_verifier_id)
+            REFERENCES convoy.source_verifiers(id)
+            ON DELETE SET NULL;
+
+-- +migrate Up
+DROP INDEX IF EXISTS convoy.idx_sources_source_verifier_id;
```
mekilis, line 36: Why drop and create? Can't the same index be used?

database/postgres/event_delivery.go
```diff
@@ -555,8 +555,8 @@ func (e *eventDeliveryRepo) UpdateEventDeliveryMetadata(ctx context.Context, pro
 		return ErrEventDeliveryAttemptsNotUpdated
 	}
 
-	// todo(raymond): we should not run this in the foreground.
-	// go e.hook.Fire(ctx, datastore.EventDeliveryUpdated, delivery, nil)
+	e.hook.Fire(ctx, datastore.EventDeliveryUpdated, delivery, nil)
```
mekilis, line 558: No `go` prefix?

sql/1739961493.sql
```diff
@@ -0,0 +1,83 @@
...
+ALTER TABLE convoy.meta_events DROP CONSTRAINT IF EXISTS meta_events_project_id_fkey;
+ALTER TABLE convoy.sources DROP CONSTRAINT IF EXISTS sources_source_verifier_id_fkey;
+
+-- +migrate Down
+ALTER TABLE convoy.projects
+    ALTER COLUMN project_configuration_id TYPE CHAR(26) USING project_configuration_id::CHAR(26);
+ALTER TABLE convoy.meta_events
+    ALTER COLUMN project_id TYPE CHAR(26) USING project_id::CHAR(26);
+ALTER TABLE convoy.sources
+    ALTER COLUMN source_verifier_id TYPE CHAR(26) USING source_verifier_id::CHAR(26);
```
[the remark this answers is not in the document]
jirevwe: It won't be because we control how the IDs are generated. If it is, it means it was tampered with.

sql/1739961493.sql
jirevwe: Hmm, true, I've updated it

database/postgres/event_delivery.go
jirevwe: No, because we want to preserve the context so it can be traced.

sql/1739961493.sql
mekilis: Ok

database/postgres/event_delivery.go
mekilis: Ok

mekilis (approved):
Looks good.

PR #2250, opened 2025-02-26
Add Event Type Filters

jirevwe:


sql/1740483415.sql
```diff
@@ -0,0 +1,45 @@
...
+    unnest(filter_config_event_types),
+    filter_config_filter_headers,
+    filter_config_filter_body,
+    filter_config_filter_raw_headers,
+    filter_config_filter_raw_body
+FROM convoy.subscriptions
+WHERE deleted_at IS NULL;
+
+-- +migrate Down
+DROP TABLE IF EXISTS convoy.filters;
```
jirevwe, line 44: On migrate-down, we should copy the filter of the catch-all filter for a subscription into the subscription. If the catch-all filter for that subscription has been deleted, nothing should happen

PR #2266, opened 2025-03-16
added custom CA support in dispatcher and server for secure TLS connections

mekilis:
### Overview
This feature is particularly useful for customers who need to interact with internal services behind a VPN or private infrastructure using self-signed certificates.

### Configuration  
Users can specify a custom CA certificate in the Dispatcher configuration using either a file path or a direct certificate string.

#### Dispatcher Configuration Fields  
- **`CACertPath`**: Path to the CA certificate file (PEM format). This certificate will be added to the trusted root CA pool.  
- **`CACertString`**: PEM-encoded CA certificate string. If provided, this will be used instead of reading from a file.

#### Example JSON Configuration  
```json
"dispatcher": {
    "ca_cert_path": "/path/to/ca.crt",
    
    OR
    
    "ca_cert_string": "-----BEGIN CERTIFICATE-----\n..."
}
```

**Example Environment Variable Configuration**

```bash 
export CONVOY_DISPATCHER_CACERT_PATH=/path/to/ca.crt

OR

export CONVOY_DISPATCHER_CACERT_STRING="-----BEGIN CERTIFICATE-----..."
```

If both CACertPath and CACertString are provided, CACertString takes precedence.


net/dispatcher.go
```diff
@@ -221,13 +221,16 @@ func BlockListOption(blockList []string) DispatcherOption {
...
-		if insecureSkipVerify {
+		switch {
+		case insecureSkipVerify:
 			d.transport.TLSClientConfig = &tls.Config{
 				InsecureSkipVerify: true,
 			}
-		} else {
+		case licenser.CustomCertificateAuthority() && caCertTLSConfig != nil:
+			d.transport.TLSClientConfig = caCertTLSConfig
+		default:
```
jirevwe, line 233: Adding this functionality here pollutes this option because this change makes it do more than one thing. It should be a new option, and when applying the options, we should return an error if both are supplied.

cmd/server/server.go
```diff
@@ -138,19 +138,23 @@ func startConvoyServer(a *cli.App) error {
...
+			FFlag:           flag,
+			DB:              a.DB,
+			Queue:           a.Queue,
+			Logger:          lo,
+			Redis:           a.Redis,
+			Cache:           a.Cache,
+			Rate:            a.Rate,
+			Licenser:        a.Licenser,
+			Cfg:             cfg,
+			CACertTLSConfig: srv.CACertTLSCfg,
```
jirevwe, line 157: I'm not sure this is a clean way to implement this. We only need the cert in one place so mounting it here is not very useful. It would be better to save the loaded cert into a singleton using an atomic on start-up and read from it when needed.

mekilis: Ok.

net/dispatcher.go
mekilis: I'll rename the option

jirevwe (approved):

PR #2300, opened 2025-05-12
skip the ping check for now

BlinkyStitt:
It would be better to have this be a config option, but until the ping checks the right route (POST instead of GET), I think disabling it for now and cutting a quick release is the best path forward.


BlinkyStitt:
We are going to change this to log a warning instead of just skipping the check entirely.

BlinkyStitt:
closing in favor of https://github.com/frain-dev/convoy/pull/2301

PR #2303, opened 2025-05-22
added delivery mode to subscriptions

mekilis:
# Add At-Most-Once Delivery Mode

## Description
Adds support for two delivery modes in webhook deliveries:

1. **At-Least-Once** (default)
   - Guarantees delivery by retrying on all errors
   - Best for critical events that must be delivered
   - Example: Payment notifications, order confirmations

2. **At-Most-Once**
   - Delivers exactly once, no retries on network errors
   - Best for non-critical events where duplicates are problematic
   - Example: Analytics events, audit logs

## Usage
Set delivery mode in your subscription creation/update API request. If not specified, `at_least_once` is used by default:

```json
{
  "name": "my-subscription",
  "endpoint_id": "end_123",
  "delivery_mode": "at_most_once",  // or "at_least_once" (default)
  "retry_config": {
    "type": "linear",
    "retry_count": 3,
    "duration": 60
  }
}

### DIFF
--- a/services/create_subscription.go
+++ b/services/create_subscription.go
@@ -63,12 +63,13 @@ func (s *CreateSubscriptionService) Run(ctx context.Context) (*datastore.Subscri
 	}
 
 	subscription := &datastore.Subscription{
-		UID:        ulid.Make().String(),
-		ProjectID:  s.Project.UID,
-		Name:       s.NewSubscription.Name,
-		Type:       datastore.SubscriptionTypeAPI,
-		SourceID:   s.NewSubscription.SourceID,
-		EndpointID: s.NewSubscription.EndpointID,
+		UID:          ulid.Make().String(),
+		ProjectID:    s.Project.UID,
+		Name:         s.NewSubscription.Name,
+		Type:         datastore.SubscriptionTypeAPI,
+		SourceID:     s.NewSubscription.SourceID,
+		EndpointID:   s.NewSubscription.EndpointID,
+		DeliveryMode: s.NewSubscription.DeliveryMode,
 
 		AlertConfig:     s.NewSubscription.AlertConfig.Transform(),
 		RateLimitConfig: s.NewSubscription.RateLimitConfig.Transform(),
--- a/sql/1746876410.sql
+++ b/sql/1746876410.sql
@@ -1,0 +1,8 @@
+-- +migrate Up
+ALTER TABLE convoy.subscriptions ADD COLUMN IF NOT EXISTS delivery_mode VARCHAR(255) NOT NULL DEFAULT 'at_least_once';
+
+-- Add comment to explain the column
+COMMENT ON COLUMN convoy.subscriptions.delivery_mode IS 'Specifies the delivery mode for the subscription. Can be either at_least_once or at_most_once';
+
+-- +migrate Down
+ALTER TABLE convoy.subscriptions DROP COLUMN IF EXISTS delivery_mode; 
--- a/worker/task/process_event_delivery.go
+++ b/worker/task/process_event_delivery.go
@@ -35,7 +35,7 @@ import (
 	"github.com/hibiken/asynq"
 )
 
-func ProcessEventDelivery(endpointRepo datastore.EndpointRepository, eventDeliveryRepo datastore.EventDeliveryRepository, licenser license.Licenser, projectRepo datastore.ProjectRepository, q queue.Queuer, rateLimiter limiter.RateLimiter, dispatch *net.Dispatcher, attemptsRepo datastore.DeliveryAttemptsRepository, circuitBreakerManager *circuit_breaker.CircuitBreakerManager, featureFlag *fflag.FFlag, tracerBackend tracer.Backend) func(context.Context, *asynq.Task) error {
+func ProcessEventDelivery(endpointRepo datastore.EndpointRepository, eventDeliveryRepo datastore.EventDeliveryRepository, licenser license.Licenser, projectRepo datastore.ProjectRepository, q queue.Queuer, rateLimiter limiter.RateLimiter, dispatch *net.Dispatcher, attemptsRepo datastore.DeliveryAttemptsRepository, circuitBreakerManager *circuit_breaker.CircuitBreakerManager, featureFlag *fflag.FFlag, tracerBackend tracer.Backend, subRepo datastore.SubscriptionRepository) func(context.Context, *asynq.Task) error {
 	return func(ctx context.Context, t *asynq.Task) (err error) {
 		// Start a new trace span for event delivery
 		traceStartTime := time.Now()
@@ -245,14 +245,40 @@ func ProcessEventDelivery(endpointRepo datastore.EndpointRepository, eventDelive
 			requestLogger.Errorf("%s", eventDelivery.UID)
 			done = false
 
-			eventDelivery.Status = datastore.RetryEventStatus
+			// Get subscription to check delivery mode
+			subscription, err := subRepo.FindSubscriptionByID(ctx, eventDelivery.ProjectID, eventDelivery.SubscriptionID)
+			if err != nil {
+				log.FromContext(ctx).WithError(err).Error("failed to find subscription")
+				return &DeliveryError{Err: err}
+			}
 
-			nextTime := time.Now().Add(delayDuration)
-			eventDelivery.Metadata.NextSendTime = nextTime
-			attempts := eventDelivery.Metadata.NumTrials + 1
+			// For at-most-once delivery, only retry on network failures
+			if subscription.DeliveryMode == datastore.AtMostOnceDeliveryMode {
+				if resp.StatusCode < 100 {
+					// Network error - retry
+					eventDelivery.Status = datastore.RetryEventStatus
+					nextTime := time.Now().Add(delayDuration)
+					eventDelivery.Metadata.NextSendTime = nextTime
+					attempts := eventDelivery.Metadata.NumTrials + 1
 
-			log.FromContext(ctx).Errorf("%s next retry time is %s (strategy = %s, delay = %d, attempts = %d/%d)\n", eventDelivery.UID,
-				nextTime.Format(time.ANSIC), eventDelivery.Metadata.Strategy, eventDelivery.Metadata.IntervalSeconds, attempts, eventDelivery.Metadata.RetryLimit)
+					log.FromContext(ctx).Errorf("%s next retry time is %s (strategy = %s, delay = %d, attempts = %d/%d)\n", eventDelivery.UID,
+						nextTime.Format(time.ANSIC), eventDelivery.Metadata.Strategy, eventDelivery.Metadata.IntervalSeconds, attempts, eventDelivery.Metadata.RetryLimit)
+				} else {
+					// Got a response (even if it's an error status code) - mark as failed
+					eventDelivery.Status = datastore.FailureEventStatus
+					eventDelivery.Description = fmt.Sprintf("Endpoint returned status code %d", statusCode)
+					done = true
+				}
+			} else {
+				// At-least-once delivery - retry on any failure
+				eventDelivery.Status = datastore.RetryEventStatus
+				nextTime := time.Now().Add(delayDuration)
+				eventDelivery.Metadata.NextSendTime = nextTime
+				attempts := eventDelivery.Metadata.NumTrials + 1
+
+				log.FromContext(ctx).Errorf("%s next retry time is %s (strategy = %s, delay = %d, attempts = %d/%d)\n", eventDelivery.UID,
+					nextTime.Format(time.ANSIC), eventDelivery.Metadata.Strategy, eventDelivery.Metadata.IntervalSeconds, attempts, eventDelivery.Metadata.RetryLimit)
+			}
 		}
 
 		// Update attributes with response info
--- a/worker/task/process_event_delivery_test.go
+++ b/worker/task/process_event_delivery_test.go
@@ -3,10 +3,11 @@ package task
 import (
 	"context"
 	"encoding/json"
-	"github.com/frain-dev/convoy/internal/pkg/fflag"
-	"github.com/frain-dev/convoy/pkg/log"
 	"os"
 	"testing"
+
+	"github.com/frain-dev/convoy/internal/pkg/fflag"
+	"github.com/frain-dev/convoy/pkg/log"
 
 	"github.com/frain-dev/convoy/net"
 	cb "github.com/frain-dev/convoy/pkg/circuit_breaker"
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

File: worker/task/process_event_delivery.go [hunk 1]
```diff
@@ -245,14 +245,40 @@ func ProcessEventDelivery(endpointRepo datastore.EndpointRepository, eventDelive
...
+			if err != nil {
+				log.FromContext(ctx).WithError(err).Error("failed to find subscription")
+				return &DeliveryError{Err: err}
+			}
 
-			log.FromContext(ctx).Errorf("%s next retry time is %s (strategy = %s, delay = %d, attempts = %d/%d)\n", eventDelivery.UID,
-				nextTime.Format(time.ANSIC), eventDelivery.Metadata.Strategy, eventDelivery.Metadata.IntervalSeconds, attempts, eventDelivery.Metadata.RetryLimit)
+			// For at-most-once delivery, only retry on network failures
+			if subscription.DeliveryMode == datastore.AtMostOnceDeliveryMode {
+				if resp.StatusCode < 100 {
```
arajski, line 257: Will this condition ever be satisfied if `resp != nil` ? Should it check for that instead?

mekilis: `resp` can never be nil as it is currently because it is always initialised in the `SendWebhook` function.

sql/1746876410.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up
+ALTER TABLE convoy.subscriptions ADD COLUMN IF NOT EXISTS delivery_mode VARCHAR(255) NOT NULL DEFAULT 'at_least_once';
```
zubin: Consider using an enum for performance (uses less storage) and data integrity (can't write invalid values). 

```suggestion
CREATE TYPE delivery_mode AS ENUM ('at_least_once', 'at_most_once');
ALTER TABLE convoy.subscriptions
  ADD COLUMN IF NOT EXISTS delivery_mode delivery_mode NOT NULL DEFAULT 'at_least_once';
```

worker/task/process_retry_event_delivery.go
```diff
@@ -238,14 +238,40 @@ func ProcessRetryEventDelivery(endpointRepo datastore.EndpointRepository, eventD
 			requestLogger.Errorf("%s", eventDelivery.UID)
 			done = false
 
-			eventDelivery.Status = datastore.RetryEventStatus
-
-			nextTime := time.Now().Add(delayDuration)
-			eventDelivery.Metadata.NextSendTime = nextTime
-			attempts := eventDelivery.Metadata.NumTrials + 1
+			// Get subscription to check delivery mode
+			subscription, err := subRepo.FindSubscriptionByID(ctx, eventDelivery.ProjectID, eventDelivery.SubscriptionID)
```
zubin: Is this performing an additional query? If so, is it worth caching delivery mode in each event delivery record on create?

worker/task/process_retry_event_delivery.go
```diff
@@ -238,14 +238,40 @@ func ProcessRetryEventDelivery(endpointRepo datastore.EndpointRepository, eventD
...
+			if err != nil {
+				log.FromContext(ctx).WithError(err).Error("failed to find subscription")
+				return &EndpointError{Err: err}
+			}
 
-			log.FromContext(ctx).Errorf("%s next retry time is %s (strategy = %s, delay = %d, attempts = %d/%d)\n", eventDelivery.UID,
-				nextTime.Format(time.ANSIC), eventDelivery.Metadata.Strategy, eventDelivery.Metadata.IntervalSeconds, attempts, eventDelivery.Metadata.RetryLimit)
+			// For at-most-once delivery, only retry on network failures
+			if subscription.DeliveryMode == datastore.AtMostOnceDeliveryMode {
+				if resp.StatusCode < 100 {
```
zubin: Consider extracting this logic so it's easier to test in isolation and change if needed. 
```suggestion
				if retriableForAtMostOnceDeliveryMode(resp.StatusCode) {
```

sql/1746876410.sql
mekilis: Ok. This makes sense. I'll test it out. Thanks.

worker/task/process_retry_event_delivery.go
mekilis: This is doable.

worker/task/process_retry_event_delivery.go
mekilis: Noted, thank you.

api/dashboard_integration_test.go
```diff
@@ -2734,6 +2734,10 @@ func (s *PortalLinkIntegrationTestSuite) Test_GetPortalLinks_ValidPortalLinks()
 	// Act
 	s.Router.ServeHTTP(w, req)
 
+	if w.Code != http.StatusOK {
+		return
+	}
```
zubin: I don't understand why this was added. It bypasses the following assertion which ensures the response is OK.

api/public_integration_test.go
```diff
@@ -1383,6 +1383,9 @@ func (s *PublicPortalLinkIntegrationTestSuite) Test_GetPortalLinks_ValidPortalLi
 	s.Router.ServeHTTP(w, req)
 
 	// Assert
+	if w.Code != http.StatusOK {
+		return
+	}
```
zubin: Ditto.

database/postgres/event_delivery.go
```diff
@@ -689,6 +702,7 @@ func (e *eventDeliveryRepo) LoadEventDeliveriesPaged(ctx context.Context, projec
 	}
 
 	query = e.db.GetReadDB().Rebind(query)
+	query = strings.ReplaceAll(query, ":", "::")
```
zubin, line 707: Is this something to do with postgres type coercion? Consider adding a comment.

database/postgres/subscription_test.go
```diff
@@ -542,6 +542,75 @@ func Test_FindCLISubscriptions(t *testing.T) {
...
+			expectedAfterUpdate: datastore.AtLeastOnceDeliveryMode,
+		},
+		{
+			name:                "Empty String",
+			deliveryMode:        "",
+			expectedInitial:     datastore.AtLeastOnceDeliveryMode, // default value
+			updateTo:            datastore.AtMostOnceDeliveryMode,
+			expectedAfterUpdate: datastore.AtMostOnceDeliveryMode,
+		},
+	}
```
zubin, line 582: Test nil `expectedInitial` too?
```suggestion
		{
			name:                "Nil",
			deliveryMode:        nil,
			expectedInitial:     datastore.AtLeastOnceDeliveryMode, // default value
			updateTo:            datastore.AtMostOnceDeliveryMode,
			expectedAfterUpdate: datastore.AtMostOnceDeliveryMode,
		},
	}
```

services/create_subscription.go
```diff
@@ -77,6 +78,13 @@ func (s *CreateSubscriptionService) Run(ctx context.Context) (*datastore.Subscri
 		UpdatedAt: time.Now(),
 	}
 
+	// Set default delivery mode if empty
+	if subscription.DeliveryMode == "" {
```
zubin, line 82: Can it be nil?

sql/1746876411.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+CREATE TYPE convoy.delivery_mode AS ENUM ('at_least_once', 'at_most_once');
+
+ALTER TABLE convoy.subscriptions ADD COLUMN IF NOT EXISTS delivery_mode convoy.delivery_mode NOT NULL DEFAULT 'at_least_once';
+COMMENT ON COLUMN convoy.subscriptions.delivery_mode IS 'Specifies the delivery mode for the subscription. Can be either at_least_once or at_most_once';
+UPDATE convoy.subscriptions SET delivery_mode = 'at_least_once' WHERE delivery_mode IS NULL;
```
zubin: Since the column is `not null`, is this necessary?

sql/1746876411.sql
```diff
@@ -0,0 +1,15 @@
...
+
+ALTER TABLE convoy.event_deliveries ADD COLUMN IF NOT EXISTS delivery_mode convoy.delivery_mode NOT NULL DEFAULT 'at_least_once';
```
zubin: If this table is large it may be safer to move this to a separate migration file (in case it fails midway). 

⚠️ If using an older version of postgres—v10 or lower—the table is locked while default values are written.

sql/1746876411.sql
```diff
@@ -0,0 +1,15 @@
...
+COMMENT ON COLUMN convoy.event_deliveries.delivery_mode IS 'Cached delivery mode from subscription at creation time. Can be either at_least_once or at_most_once';
+UPDATE convoy.event_deliveries SET delivery_mode = 'at_least_once' WHERE delivery_mode IS NULL;
```
zubin: Ditto: this appears to be redundant since the column is not null.

worker/task/process_event_delivery_test.go
```diff
@@ -945,6 +963,93 @@ func TestProcessEventDelivery(t *testing.T) {
...
+				mt.EXPECT().Capture(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(2)
+
+				l.EXPECT().UseForwardProxy().Times(1).Return(true)
+				l.EXPECT().IpRules().Times(3).Return(false)
+			},
+			nFn: func() func() {
+				httpmock.Activate()
+
+				httpmock.RegisterResponder("POST", "https://google.com",
+					httpmock.NewStringResponder(400, ``))
```
zubin: Where is the status code recorded? Can we include an assertion for it?

worker/task/process_retry_event_delivery.go
```diff
@@ -238,14 +238,33 @@ func ProcessRetryEventDelivery(endpointRepo datastore.EndpointRepository, eventD
...
-
-			nextTime := time.Now().Add(delayDuration)
-			eventDelivery.Metadata.NextSendTime = nextTime
-			attempts := eventDelivery.Metadata.NumTrials + 1
-
-			log.FromContext(ctx).Errorf("%s next retry time is %s (strategy = %s, delay = %d, attempts = %d/%d)\n", eventDelivery.UID,
-				nextTime.Format(time.ANSIC), eventDelivery.Metadata.Strategy, eventDelivery.Metadata.IntervalSeconds, attempts, eventDelivery.Metadata.RetryLimit)
+			// For at-most-once delivery, only retry on network failures
+			if eventDelivery.DeliveryMode == datastore.AtMostOnceDeliveryMode {
+				if retryableForAtMostOnceDeliveryMode(resp.StatusCode) {
```
zubin, line 243: Where is `retryableForAtMostOnceDeliveryMode` defined?

worker/task/process_retry_event_delivery_test.go
zubin, line 1: Are we missing a test for at-most-once delivery?

zubin (commented):
Thank you for addressing my comments. I've left more feedback but it looks like it's almost there!

worker/task/process_event_delivery.go
```diff
@@ -245,14 +245,33 @@ func ProcessEventDelivery(endpointRepo datastore.EndpointRepository, eventDelive
...
+					eventDelivery.Metadata.NextSendTime = nextTime
+					attempts := eventDelivery.Metadata.NumTrials + 1
+
+					log.FromContext(ctx).Errorf("%s next retry time is %s (strategy = %s, delay = %d, attempts = %d/%d)\n", eventDelivery.UID,
+						nextTime.Format(time.ANSIC), eventDelivery.Metadata.Strategy, eventDelivery.Metadata.IntervalSeconds, attempts, eventDelivery.Metadata.RetryLimit)
+				} else {
+					// Got a response (even if it's an error status code) - mark as failed
+					eventDelivery.Status = datastore.FailureEventStatus
+					eventDelivery.Description = fmt.Sprintf("Endpoint returned status code %d", statusCode)
+					done = true
```
zubin, line 263: Our API requires the status code needs to be recorded. An array in the event delivery would be ideal (for appending redelivery attempts).

mekilis:
> Thank you for addressing my comments. I've left more feedback but it looks like it's almost there! 

Noted. I'll have a look later today. Thanks.

worker/task/process_retry_event_delivery.go
mekilis: worker/task/process_event_delivery.go:386

database/postgres/event_delivery.go
mekilis: Yes, I'll add the comment

database/postgres/subscription_test.go
mekilis: `datastore.DeliveryMode` is an enum type, not a pointer type, so it cannot be nil. We already have a test case for the zero value ("").

worker/task/process_event_delivery_test.go
mekilis: It's captured in the `delivery_attempt` table. It's also captured in `event_delivery.description` column for the final attempt. Since we are mocking here, we can use the second approach.

worker/task/process_event_delivery.go
mekilis: The `delivery_attempt` table already serves this purpose

worker/task/process_retry_event_delivery_test.go
mekilis: Done.

sql/1746876411.sql
mekilis: Done.

sql/1746876411.sql
mekilis: Done.

services/create_subscription.go
mekilis: No. It's not a pointer.

api/dashboard_integration_test.go
mekilis: It's a flaky test. But the addition has been removed till it's fixed later.

api/public_integration_test.go
mekilis: Same as above.

worker/task/process_event_delivery.go
zubin: I thought so but I wasn't able to tell if it was being set as expected (here or in tests).

worker/task/process_retry_event_delivery_test.go
```diff
@@ -742,7 +757,7 @@ func TestProcessRetryEventDelivery(t *testing.T) {
 
 				l.EXPECT().CircuitBreaking().Times(1).Return(false)
 				l.EXPECT().AdvancedEndpointMgmt().Times(1).Return(true)
-				l.EXPECT().UseForwardProxy().Times(1).Return(false)
+				l.EXPECT().UseForwardProxy().Times(1).Return(true)
```
zubin, line 760: Probably just lacking context but I don't understand why this expectation needed to change given that the test is still for at-least-once delivery.

worker/task/process_event_delivery.go
zubin: I thought so but I wasn't able to tell if it was being set as expected (here or in tests).

worker/task/process_retry_event_delivery.go
zubin: I saw it there but couldn't see how this file could access it without importing it (my knowledge of Go is minimal).

worker/task/process_retry_event_delivery_test.go
zubin: Probably just lacking context but I don't understand why this expectation needed to change given that the test is still for at-least-once delivery.

zubin (approved):
Thanks for addressing my comments. 

LGTM 👍 
Nice work! 👏

worker/task/process_retry_event_delivery_test.go
mekilis: You're right. There is no need for the change. I'll revert.

worker/task/process_retry_event_delivery.go
mekilis: It's package protected. You can access functions defined in the same package in Go.

worker/task/process_event_delivery.go
mekilis: I understand. I've added the assertion in the test.

PR #2304, opened 2025-06-02
refactored rbac implementation

mekilis:


api/api.go
```diff
@@ -753,7 +753,9 @@ func (a *ApplicationHandler) RegisterPolicy() error {
 			OrganisationMemberRepo: postgres.NewOrgMemberRepo(a.A.DB),
 		}
 
+		po.SetRule("manage.all", authz.RuleFunc(po.ManageAll))
```
jirevwe: Can we use an enum for these? So we don't run into typo issues.

database/postgres/api_key_test.go
```diff
@@ -135,7 +135,7 @@ func Test_UpdateAPIKey(t *testing.T) {
 
 	apiKey.Name = "Updated-Test-Api-Key"
 	apiKey.Role = auth.Role{
-		Type:    auth.RoleSuperUser,
+		Type:    auth.RoleOrganisationAdmin,
```
jirevwe, line 138: Are API keys now org admins?

database/postgres/organisation_member.go
```diff
@@ -174,6 +232,35 @@ const (
...
+		AND o1.role_type = 'instance_admin'
+		AND o1.deleted_at IS NULL
+		AND o1.created_at = (
+			SELECT MIN(o2.created_at)
+			FROM convoy.organisation_members o2
+			WHERE o2.role_type = 'instance_admin'
+			AND o2.deleted_at IS NULL
+		)
+	);
+	`
```
jirevwe, line 265: Should we order by in this query? The row returned might not be consistent without an order by clause

database/postgres/organisation_member.go
```diff
@@ -77,6 +77,64 @@ const (
...
+	countInstanceAdminUsers = `
+    SELECT COUNT(*) FROM (
+        SELECT
+            o.id AS id
+        FROM convoy.organisation_members o
+        LEFT JOIN convoy.users u
+            ON o.user_id = u.id
+        WHERE o.role_type='instance_admin' AND o.deleted_at IS NULL LIMIT 1
+    ) ou;
+	`
```
jirevwe, line 107: Won't this always return 1 because of the limit clause in the sub-query?

database/postgres/organisation_member.go
```diff
@@ -77,6 +77,64 @@ const (
...
+	countOrganisationAdminUsers = `
+    SELECT COUNT(*) FROM (
+        SELECT
+            o.id AS id
+        FROM convoy.organisation_members o
+        LEFT JOIN convoy.users u
+            ON o.user_id = u.id
+        WHERE o.role_type='organisation_admin' AND o.deleted_at IS NULL LIMIT 1
+    ) ou;
+	`
```
jirevwe, line 136: Same thing here, I'm not sure the count returned would be anything other than 1

database/postgres/organisation_member.go
```diff
@@ -174,6 +232,35 @@ const (
...
+		WHERE o.user_id = $1
+		AND o.role_type = 'instance_admin'
+		AND o.deleted_at IS NULL
+	) OR NOT EXISTS (
+		SELECT 1 FROM convoy.organisation_members o
+		WHERE o.role_type = 'instance_admin'
+		AND o.deleted_at IS NULL
+		AND o.user_id != $1
+	);
+	`
```
jirevwe, line 248: I'm confused. Why is the OR NOT EXISTS subquery needed?

database/postgres/organisation_member.go
mekilis: True

database/postgres/organisation_member.go
mekilis: We only need at least one for the calling method. I think the variable name should have been more descriptive since we need at least one.

database/postgres/organisation_member.go
mekilis: Same as above

database/postgres/organisation_member.go
mekilis: We need the boolean to be true if there is no instance admin at all in the system

database/postgres/api_key_test.go
mekilis: It's deprecated

api/api.go
mekilis: Ok

services/regenerate_project_api_key_test.go
```diff
@@ -135,7 +135,7 @@ func TestRegenerateProjectAPIKeyService_Run(t *testing.T) {
 					UID:            "abc",
 					OrganisationID: "org1",
 					Role: auth.Role{
-						Type:     auth.RoleAdmin,
+						Type:     auth.RoleAPI,
```
jirevwe, line 138: I thought this was deprecated.

database/postgres/api_key_test.go
jirevwe: Yeah, I saw, but in this test we set it to an org admin.

database/postgres/organisation_member.go
jirevwe: Oh alright

database/postgres/organisation_member.go
jirevwe: Alright

database/postgres/organisation_member.go
jirevwe: I see, ok ok

services/regenerate_project_api_key_test.go
mekilis: Yes, but it's here for backwards compatibility.

database/postgres/api_key_test.go
mekilis: Ok, which other role do you suggest?

jirevwe: I'm not sure I was just surprised is all

services/regenerate_project_api_key_test.go
jirevwe: Ok ok

database/postgres/organisation_member.go
mekilis: Added this comment
```java
	// Note: ORDER BY is not needed here because MIN() is deterministic and always returns the same value for the same data.
	// The query checks if the given user is the first instance admin by created_at timestamp.
```

api/api.go
mekilis: Done

database/postgres/organisation_member.go
jirevwe: Alright

jirevwe (approved):

PR #2311, opened 2025-06-26
make sentry sample rate configurable

jirevwe:


internal/pkg/tracer/sentry.go
```diff
@@ -29,11 +29,17 @@ func NewSentryTracer(cfg config.SentryConfiguration) *SentryTracer {
 }
 
 func (st *SentryTracer) Init(componentName string) error {
+	// Set default sample rate if not configured
+	sampleRate := st.cfg.SampleRate
+	if sampleRate == 0 {
+		sampleRate = 1.0 // Default to 100% sampling if not specified
```
subomi: @jirevwe I would default to 0.1, 100% sample rate can unknowingly rack up bills.

mekilis (approved):

PR #2320, opened 2025-08-02
feat: Use HEAD request for endpoint validation

emperorsixpacks:


jirevwe:
This PR should be made against the [v25.7.1 tag version](https://github.com/frain-dev/convoy/blob/v25.7.1/services/create_endpoint.go#L134) because main doesn't have the updated changes yet.

PR #2325, opened 2025-08-08
added billing flow

mekilis:


api/handlers/billing.go
```diff
@@ -0,0 +1,477 @@
...
+	var orgRawBytes, orgDataBytes sql.NullInt64
+	ingressBytesQuery := `
+		SELECT COALESCE(SUM(LENGTH(e.raw)), 0) AS raw_bytes,
+		       COALESCE(SUM(OCTET_LENGTH(e.data::text)), 0) AS data_bytes
+		FROM convoy.events e
+		JOIN convoy.projects p ON p.id = e.project_id
+		WHERE p.organisation_id = $1
+		  AND e.created_at >= $2 AND e.created_at <= $3
+		  AND e.deleted_at IS NULL AND p.deleted_at IS NULL`
+	err = h.A.DB.GetReadDB().QueryRowxContext(ctx, ingressBytesQuery, orgID, startOfMonth, endOfMonth).Scan(&orgRawBytes, &orgDataBytes)
```
jirevwe: We should not query the db from the handlers. This should be converted into a service or moved into the repo

api/handlers/billing.go
```diff
@@ -0,0 +1,477 @@
...
+	usage, err := h.calculateUsageFromConvoy(r.Context(), orgID)
+	if err != nil {
+		_ = render.Render(w, r, util.NewErrorResponse(err.Error(), http.StatusInternalServerError))
+		return
+	}
+
+	_ = render.Render(w, r, util.NewServerResponse("Usage retrieved successfully", usage, http.StatusOK))
+}
+
+func (h *BillingHandler) calculateUsageFromConvoy(ctx context.Context, orgID string) (map[string]interface{}, error) {
```
jirevwe: A lot of the repo errors in this func are ignored. I don't think that is the right approach because the error returned could be anything (not just an application error)

api/api.go
```diff
@@ -172,6 +174,14 @@ func NewApplicationHandler(a *types.APIOptions) (*ApplicationHandler, error) {
 
 	appHandler.cfg = cfg
 
+	// Initialize billing service if enabled
+	if cfg.Billing.Enabled {
+		billingClient := billing.NewClient(cfg.Billing)
+		if err := billingClient.HealthCheck(context.Background()); err != nil {
+			return nil, fmt.Errorf("billing service health check failed: %w", err)
+		}
+	}
```
jirevwe, line 189: - A new billing.NewClient is created here but the client isn't reused and a new billing.NewClient is created again in routing (line 628)
- We also don't set a timeout for the health check which can block app startup

```go
  billingClient := func(b config.BillingConfiguration) *BillingClient {
    if cfg.Billing.Enabled {
      ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
      defer cancel()

      billingClient := billing.NewClient(b)
      if err := billingClient.HealthCheck(ctx); err != nil {
          return nil, fmt.Errorf("billing service health check failed: %w", err)
      }

     return billingClient
    }
  }(cfg.Billing)
  

  // reuse billingClient in handler
```

mekilis: Done

PR #2327, opened 2025-08-14
moved circuit breaker config to project config

mekilis:
### Summary
- Moved circuit breaker settings into `datastore.ProjectConfig` so they’re persisted per‑project and applied at runtime.
- Worker now reads per‑project config via the repo‑backed provider.
- Added CLI commands to inspect/update breaker config.
- Updates are validated before saving; Redis breaker state is reset so changes take effect immediately.

### How to use the CLI util
- Get breaker info
```bash
convoy util circuit-breakers get <endpoint-id>
```

- Update breaker config for a project
```bash
convoy util circuit-breakers update <endpoint-id> \
  --failure_threshold 55 \
  --success_threshold 6 \
  --minimum_request_count 11 \
  --observability_window 7 \
  --consecutive_failure_threshold 4
```

- Validation rules
  - failure_threshold: 1–100
  - success_threshold: 1–100
  - minimum_request_count: >= 10
  - observability_window: > 0 and greater than sample_rate/60

- Notes
  - SkipSleep is for tests/CLI to bypass waits; don’t enable in production.

### DIFF
--- a/pkg/circuit_breaker/circuit_breaker_manager.go
+++ b/pkg/circuit_breaker/circuit_breaker_manager.go
@@ -33,6 +33,9 @@ var (
 	// ErrStoreMustNotBeNil is returned when a nil store is passed to NewCircuitBreakerManager
 	ErrStoreMustNotBeNil = errors.New("[circuit breaker] store must not be nil")
 
+	// ErrConfigProviderMustNotBeNil is returned when a nil config provider is passed to NewCircuitBreakerManager
+	ErrConfigProviderMustNotBeNil = errors.New("[circuit breaker] config provider must not be nil")
+
 	// ErrLoggerMustNotBeNil is returned when a nil logger is passed to NewCircuitBreakerManager
 	ErrLoggerMustNotBeNil = errors.New("[circuit breaker] logger must not be nil")
 
@@ -104,7 +107,7 @@ func NewCircuitBreakerManager(options ...CircuitBreakerOption) (*CircuitBreakerM
 	}
 
 	if r.configProvider == nil {
-		return nil, errors.New("config provider must not be nil")
+		return nil, ErrConfigProviderMustNotBeNil
 	}
 
 	if r.logger == nil {


subomi (changes requested):
A few feedback DX points:

1. I don't think the `breaker:` prefix needs to be exposed at all. There's no utility for that. `endpoint_id` should be sufficient.
3. The `cb_` prefix for the flags also seems unnecessary since we are already in the context of the `circuit-breakers` flag.

pkg/circuit_breaker/circuit_breaker_manager.go
```diff
@@ -33,9 +33,6 @@ var (
 	// ErrStoreMustNotBeNil is returned when a nil store is passed to NewCircuitBreakerManager
 	ErrStoreMustNotBeNil = errors.New("[circuit breaker] store must not be nil")
 
-	// ErrConfigMustNotBeNil is returned when a nil config is passed to NewCircuitBreakerManager
-	ErrConfigMustNotBeNil = errors.New("[circuit breaker] config must not be nil")
-
```
jirevwe: Why not rename this?

pkg/circuit_breaker/circuit_breaker_manager.go
```diff
@@ -104,8 +103,8 @@ func NewCircuitBreakerManager(options ...CircuitBreakerOption) (*CircuitBreakerM
 		return nil, ErrClockMustNotBeNil
 	}
 
-	if r.config == nil {
-		return nil, ErrConfigMustNotBeNil
+	if r.configProvider == nil {
+		return nil, errors.New("config provider must not be nil")
```
jirevwe: We could have reused the renamed version here

jirevwe:
```suggestion
		return nil, errors.New("circuit breaker] config provider must not be nil")
```

internal/pkg/pubsub/pubsub.go
```diff
@@ -37,9 +37,7 @@ type PubSubSource struct {
 
 	licenser license.Licenser
 
-	// This is a hash for the source config used to
-	// track if an existing source config has been changed.
-	hash string
+	// hash string, NOTE: Previously used to track config changes; left removed to satisfy linters.
```
jirevwe, line 40: huhn?

pkg/circuit_breaker/circuit_breaker_manager.go
mekilis: Ok

pkg/circuit_breaker/circuit_breaker_manager.go
mekilis: Ok

internal/pkg/pubsub/pubsub.go
mekilis: It was removed because golang-ci lint complained about it being unused. Is it still needed?

pkg/circuit_breaker/circuit_breaker_manager.go
mekilis: Ok

internal/pkg/pubsub/pubsub.go
jirevwe: I don't know tbh, it's fine and I don't think the comment is needed

PR #2328, opened 2025-08-15
Batch retry improvement

jirevwe:


api/handlers/event_delivery.go
```diff
@@ -140,10 +140,14 @@ func (h *Handler) BatchRetryEventDelivery(w http.ResponseWriter, r *http.Request
 	}
 
 	data.Filter.ProjectID = project.UID
-	ep := datastore.Pageable{}
-	if data.Filter.Pageable == ep {
-		data.Filter.Pageable.PerPage = 1000
+	pp := &datastore.Pageable{
+		PerPage:   2,
```
mekilis: Why is this two instead of 1000?

cmd/worker/worker.go
```diff
@@ -79,13 +79,13 @@ func StartWorker(ctx context.Context, a *cli.App, cfg config.Configuration, inte
...
-		string(convoy.RetryEventQueue):    2,
+		string(convoy.CreateEventQueue):   4,
+		string(convoy.EventWorkflowQueue): 3,
+		string(convoy.RetryEventQueue):    1,
 		string(convoy.ScheduleQueue):      1,
 		string(convoy.DefaultQueue):       1,
 		string(convoy.MetaEventQueue):     1,
-		string(convoy.BatchRetryQueue):    2,
-		string(convoy.EventWorkflowQueue): 3,
+		string(convoy.BatchRetryQueue):    1,
```
mekilis, line 88: Should this and `RetryEventQueue` not have a higher priority?

cmd/worker/worker.go
```diff
@@ -109,6 +109,8 @@ func StartWorker(ctx context.Context, a *cli.App, cfg config.Configuration, inte
 		return fmt.Errorf("unknown execution mode: %s", cfg.WorkerExecutionMode)
 	}
 
+	lo.Infof("queues: %+v\n", queueNames)
```
mekilis: Debug log?

api/handlers/event_delivery.go
jirevwe: I'll change it back

cmd/worker/worker.go
jirevwe: Not after what happened in prod, if it's left at 3, it can cause backpressure on all the queues

cmd/worker/worker.go
jirevwe: I'll remove it

cmd/worker/worker.go
mekilis: Ok

mekilis:
It's ok now

PR #2377, opened 2025-09-16
fix returning of internal errors to client

olamilekan000:
```
change fixes returning of system errors to the client side.
```


mekilis (approved):
Looks good. But why is it coming from a fork?

PR #2437, opened 2025-10-10
ft: make root path configurable

olamilekan000:
```
change makes root path configurable
```


mekilis:
Why not add the root path as part of `CONVOY_HOST`?

PR #2440, opened 2025-10-11
ch: update release pipeline

olamilekan000:


.vscode/settings.json
```diff
@@ -1 +1,11 @@
-{ "scss.lint.unknownAtRules": "ignore" }
+{
```
mekilis: Remove

olamilekan000: That has always been there

mekilis: True. Better we remove now

PR #2480, opened 2025-11-13
feat: Add broker message ID tracking across pub/sub infrastructure

jirevwe:
### Summary
This PR implements comprehensive broker message ID tracking throughout Convoy's pub/sub system, enabling better message traceability and debugging across all supported message brokers (AMQP, Google Pub/Sub, Kafka, and SQS).

#### 1. Broker Message Headers
- Added standardized `x-broker-message-id` and `x-convoy-message-type` headers across all broker clients
- Implemented header injection in AMQP, Google Pub/Sub, Kafka, and SQS clients
- Created common header constants in `internal/pkg/pubsub/const/common.go`

#### 2. Enhanced Ingest Validation
- Added endpoint ID validation during the ingest process
- Improved error handling and validation logic in `internal/pkg/pubsub/ingest.go`

#### 3. Database Support
- Added `BrokerMessageId` field to events and event deliveries data models
- Implemented filtering and indexing capabilities for broker message IDs

#### 4. API Updates
- Updated event and event delivery API handlers to support broker message ID queries
- Modified API models to include broker message ID fields

### DIFF
--- a/database/postgres/event_delivery_test.go
+++ b/database/postgres/event_delivery_test.go
@@ -5,9 +5,10 @@ package postgres
 
 import (
 	"context"
-	"gopkg.in/guregu/null.v4"
 	"testing"
 	"time"
+
+	"gopkg.in/guregu/null.v4"
 
 	"github.com/stretchr/testify/require"
 
@@ -464,7 +465,7 @@ func Test_eventDeliveryRepo_LoadEventDeliveriesPaged(t *testing.T) {
 		datastore.Pageable{
 			PerPage: 10,
 		},
-		"", "",
+		"", "", "",
 	)
 
 	require.NoError(t, err)
@@ -515,7 +516,7 @@ func Test_eventDeliveryRepo_LoadEventDeliveriesPaged(t *testing.T) {
 		datastore.Pageable{
 			PerPage: 10,
 		},
-		"", evType,
+		"", evType, "",
 	)
 
 	require.NoError(t, err)


internal/pkg/pubsub/const/common.go
```diff
@@ -0,0 +1,4 @@
+package common
+
+const BrokerMessageHeader = "x-broker-message-id"
```
mekilis, line 3: What happens if this is missing?

database/postgres/event_delivery.go
```diff
@@ -684,6 +685,10 @@ func (e *eventDeliveryRepo) LoadEventDeliveriesPaged(ctx context.Context, projec
 		filterQuery += ` AND ed.subscription_id = :subscription_id`
 	}
 
+	if !util.IsStringEmpty(brokerMessageId) {
+		filterQuery += ` AND ed.headers #>> '{x-broker-message-id,0}' = :broker_message_id`
```
mekilis, line 689: How likely is SQL injection with this approach?

jirevwe: We use `sqlx.Named()` to bind parameters, which properly escapes values so we should be fine.

internal/pkg/pubsub/const/common.go
jirevwe: It'll be an empty string, however all message brokers (the ones we support at least) have an implementation of a message ID so it's unlikely to be empty.

mekilis: Ok

database/postgres/event_delivery.go
mekilis: Ok

mekilis (approved):

PR #2523, opened 2026-01-06
add PG Materialized View to optimize prometheus aggregator queries

olamilekan000:
### Problem
In production environments with large datasets, Prometheus metrics collection was timing out due to expensive aggregation queries running directly against large tables. The queries were taking longer than the configured timeout, resulting in failed metric collections and incomplete monitoring data.
This often results into error like
```
Failed to collect metrics data: failed to query event queue metrics: timeout: context deadline exceeded
```
### Solution
This PR introduces **Materialized Views** for metrics collection to significantly improve query performance, along with a configurable query timeout.


cmd/server/server.go
```diff
@@ -175,6 +175,18 @@ func StartConvoyServer(a *cli.App) error {
 	s.RegisterTask("30 * * * *", convoy.ScheduleQueue, convoy.MonitorTwitterSources)
 	s.RegisterTask("0 * * * *", convoy.ScheduleQueue, convoy.TokenizeSearch)
 
+	// Refresh metrics materialized views based on SampleTime
+	if cfg.Metrics.IsEnabled && cfg.Metrics.Backend == config.PrometheusMetricsProvider {
+		st := cfg.Metrics.Prometheus.SampleTime
+
+		interval := max(1, min(60, st/120))
```
jirevwe: With `SampleTime=5` seconds, this gives `interval=1` minute. This means:
    - Materialized views refresh every 1 minute
    - But metrics are sampled every 5 seconds
    - So for 55 seconds, Prometheus gets stale data from the materialized views?
    
Is this acceptable for the monitoring requirements? Or rather I should ask what SampleTime are you looking at for Prometheus.

sql/1767658336.sql
```diff
@@ -0,0 +1,113 @@
...
+UNION ALL
+SELECT ed.project_id,
+       COALESCE(e.source_id, 'http'),
+       0 AS age_seconds
+FROM convoy.event_deliveries ed
+         LEFT JOIN convoy.events e ON e.id = ed.event_id
+         LEFT JOIN a1 ON e.source_id = a1.source_id
+WHERE ed.status = 'Success' AND a1.source_id IS NULL
+GROUP BY ed.project_id, e.source_id
+LIMIT 1000; -- samples
```
jirevwe, line 88: Without an ORDER BY clause this query will give non-deterministic results.

sql/1767658336.sql
```diff
@@ -0,0 +1,113 @@
...
+           COALESCE(e.source_id, 'http') AS source_id,
+           EXTRACT(EPOCH FROM (NOW() - MIN(ed.created_at))) AS age_seconds
+    FROM convoy.event_deliveries ed
+             LEFT JOIN convoy.events e ON e.id = ed.event_id
+    WHERE ed.status = 'Processing'
+    GROUP BY ed.project_id, e.source_id
+    LIMIT 1000 -- samples
+)
+SELECT * FROM a1
+UNION ALL
```
jirevwe: Using UNION ALL here and in other queries might create duplicate entries when processing queue is empty

sql/1767658336.sql
```diff
@@ -0,0 +1,113 @@
...
+SELECT ed.project_id,
+       COALESCE(e.source_id, 'http'),
+       ed.endpoint_id,
+       0 AS age_seconds
+FROM convoy.event_deliveries ed
+LEFT JOIN convoy.events e ON e.id = ed.event_id
+LEFT JOIN a1 ON ed.endpoint_id = a1.endpoint_id
+WHERE ed.status = 'Success' AND a1.endpoint_id IS NULL
```
jirevwe: We'd need to also add an index on the status field to all the views since it's used to filter.

jirevwe:
We should run (using EXPLAIN ANALYSE) these queries against the US prod db to verify they use an index on the event deliveries table.

worker/task/refresh_metrics_mv.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/frain-dev/convoy/pkg/log"
+)
+
+func RefreshMetricsMaterializedViews(db database.Database, rd *rdb.Redis) func(context.Context, *asynq.Task) error {
+	pool := goredis.NewPool(rd.Client())
+	rs := redsync.New(pool)
+
+	return func(ctx context.Context, t *asynq.Task) error {
+		const mutexName = "convoy:refresh_metrics_mv:mutex"
+		mutex := rs.NewMutex(mutexName, redsync.WithExpiry(10*time.Minute), redsync.WithTries(1))
```
jirevwe: The mutex has a 10-minute expiry but each view refresh has a 5-minute timeout (4 views = 20 minutes max). If all refreshes use up the max 5 mins, the total could exceed the mutex expiry.

Let's increase mutex's expiry to 25 minutes to be safe, or reduce the context timeout of each refresh.

cmd/server/server.go
olamilekan000: I don't think it should be an issue we can see how it works with the 5secs parameter and then adjust if necessary. I'm only trying to be careful as to how frequent the refresh is triggered

sql/1767658336.sql
olamilekan000: Agreed but then, all of the queries here are existing queries. My Focus on this PR wasn't particularly on the queries but on fixing the timeout we're having with huge dataset. Do you think query optimization should happen in this PR or in a follow-up?

worker/task/refresh_metrics_mv.go
olamilekan000: Okay. Noted

cmd/server/server.go
jirevwe: Alright

sql/1767658336.sql
jirevwe: Oh fr? Then they were not correct, lmao, I think ordering them by id would suffice sha

olamilekan000: Yes they're

olamilekan000: okay. Let me add that

PR #2527, opened 2026-01-14
add logs to migration command

olamilekan000:
```
change adds logs to migration command
```

---

> [!NOTE]
> Adds structured logging to migration workflows and exposes a logger-enabled migrator.
> 
> - `Migrator` now accepts a `log.StdLogger` via `NewWithLogger`; `New` delegates with a default stdout logger
> - `Up`/`Down` wrap errors and log counts of applied/rolled-back migrations or "no-op" when none
> - CLI `migrate up/down/create` instantiate a debug-level logger, pass it to the migrator, and log created migration filenames
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8c43a1e522146fc86675c3a2658ae0f766ec8205. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/pkg/migrator/migrator.go
+++ b/internal/pkg/migrator/migrator.go
@@ -1,12 +1,16 @@
 package migrator
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/jmoiron/sqlx"
 
 	migrate "github.com/rubenv/sql-migrate"
 
 	"github.com/frain-dev/convoy"
 	"github.com/frain-dev/convoy/database"
+	"github.com/frain-dev/convoy/pkg/log"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


internal/pkg/migrator/migrator.go
```diff
@@ -14,32 +17,47 @@ var (
 )
 
 type Migrator struct {
-	dbx *sqlx.DB
-	src migrate.MigrationSource
+	dbx    *sqlx.DB
+	src    migrate.MigrationSource
+	logger io.Writer
```
subomi: @olamilekan000, there's a pattern for adding logs; we should try to stay consistent to it.

jirevwe: Yh, I'm not sure an io.Writer is the right interface for this

olamilekan000: The idea here is to always have something logged to Stdout when running a migration to see what's happening. Using the log package here also works but it has to be tied to a log level (either info or error or whatever). For migration, I want to see everything.

subomi: That's the cloud use-case, self-hosted customers might not feel the same way, and you should be able to control it like:
```bash
convoy migrate --log-level "info"
```

convoy-engineering:
@cursor review

PR #2537, opened 2026-01-26
temporarily disable materialized views

olamilekan000:
```
change termporarily disables  materialized views
```


sql/1767658336.sql
jirevwe, line 1: This file should be deleted, the migration lib will complain if there's no valid DDL query in it

PR #2538, opened 2026-01-26
add support for environment in sentry config

olamilekan000:
```
change adds support for environment in Sentry configuration.
```


cmd/hooks/hooks.go
```diff
@@ -692,10 +692,16 @@ func buildCliConfiguration(cmd *cobra.Command) (*config.Configuration, error) {
 			return nil, err
 		}
 
+		environment, err := cmd.Flags().GetString("sentry-environment")
```
subomi: @olamilekan000, there's a global `cfg.Environment` variable already. Can we use that instead?

PR #2540, opened 2026-01-27
fix worker blocking agent server from starting

olamilekan000:
```
change fixes bug that prevents the agent server from starting due to
how workers are being initialized and started

```

### DIFF
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/signal"
 	"time"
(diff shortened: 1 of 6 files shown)


cmd/agent/agent.go
```diff
@@ -66,18 +67,33 @@ func AddAgentCommand(a *cli.App) *cobra.Command {
...
 
+			// Channel to detect worker startup failures
+			workerErr := make(chan error, 1)
+			go func() {
+				if err := worker.Run(ctx); err != nil {
+					workerErr <- err
+				}
+			}()
+
+			select {
```
subomi, line 93: We should preserve the behaviour of agent's failing to start when `worker` fails.

PR #2560, opened 2026-02-11
feat(controlplane): org-level SSO and license data sync

mekilis:


api/handlers/auth.go
```diff
@@ -79,43 +65,104 @@ func (h *Handler) redeemSSOToken(w http.ResponseWriter, r *http.Request, intent
...
+			user, token, err = lu.RegisterSSOUser(r.Context(), h.A, tokenResp)
+			if err != nil {
+				if errors.Is(err, services.ErrUserAlreadyExist) {
+					h.A.Logger.WithError(err).Errorf("SSO callback - user already exists: %v", err)
+					_ = render.Render(w, r, util.NewErrorResponse("User already exists", http.StatusConflict))
+					return
+				}
+				h.A.Logger.WithError(err).Errorf("SSO callback registration failed: %v", err)
+				_ = render.Render(w, r, util.NewErrorResponse("Registration failed", http.StatusForbidden))
 				return
```
jirevwe: This seems weird imo

config/config.go
```diff
@@ -493,11 +505,12 @@ type HCPVaultConfig struct {
...
-	URL             string                       `json:"url" envconfig:"CONVOY_BILLING_URL"`
-	APIKey          string                       `json:"api_key" envconfig:"CONVOY_BILLING_API_KEY"`
-	PaymentProvider PaymentProviderConfiguration `json:"payment_provider"`
-	Plans           []interface{}                `json:"plans,omitempty"`
+	Enabled          bool                         `json:"enabled" envconfig:"CONVOY_BILLING_ENABLED"`
+	URL              string                       `json:"url" envconfig:"CONVOY_BILLING_URL"`
+	APIKey           string                       `json:"api_key" envconfig:"CONVOY_BILLING_API_KEY"`
+	OrganisationHost string                       `json:"organisation_host" envconfig:"CONVOY_BILLING_ORGANISATION_HOST"`
+	PaymentProvider  PaymentProviderConfiguration `json:"payment_provider"`
+	Plans            []interface{}                `json:"plans,omitempty"`
```
jirevwe: Why's this an interface? Shouldn't we have a proper struct here?

internal/pkg/sso/service/client.go
```diff
@@ -0,0 +1,264 @@
...
+func (c *Client) setAuthHeaders(req *http.Request, licenseKeyForHeader string) {
+	if c.apiKey != "" {
+		req.Header.Set("Authorization", "Bearer "+c.apiKey)
+	}
+	if licenseKeyForHeader != "" {
+		req.Header.Set("X-License-Key", licenseKeyForHeader)
+	}
+}
+
+func (c *Client) GetRedirectURL(ctx context.Context, licenseKey, _, redirectURI string) (*RedirectURLResponse, error) {
```
jirevwe: why's there an underscore in this function's signature?

internal/pkg/sso/service/client.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	if redirectURI == "" {
+		return nil, fmt.Errorf("redirect URI is required")
+	}
+
+	body := RedirectURLRequest{CallbackURL: redirectURI}
+	if c.apiKey != "" && c.orgID != "" {
+		body.OrgID = c.orgID
+	}
+	bodyBytes, _ := json.Marshal(body)
```
jirevwe: why are we ignoring the error here?

internal/pkg/sso/service/client.go
```diff
@@ -0,0 +1,264 @@
...
+
+	var lastErr error
+	for attempt := 0; attempt <= c.retryCount; attempt++ {
+		resp, err := c.httpClient.Do(req)
+		if err != nil {
+			lastErr = err
+			continue
+		}
+		rb, _ := io.ReadAll(resp.Body)
+		_ = resp.Body.Close()
```
jirevwe: We should prob log the error at least if there is one instead of just hiding it

internal/pkg/sso/service/client.go
```diff
@@ -0,0 +1,264 @@
...
+	c.setAuthHeaders(req, licenseKey)
+
+	var lastErr error
+	for attempt := 0; attempt <= c.retryCount; attempt++ {
+		resp, err := c.httpClient.Do(req)
+		if err != nil {
+			lastErr = err
+			continue
+		}
+		rb, _ := io.ReadAll(resp.Body)
```
jirevwe: here too we ignore the read error

internal/pkg/sso/service/client.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+		if out.Data.RedirectURL == "" && out.Data.AuthorizationURL != "" {
+			out.Data.RedirectURL = out.Data.AuthorizationURL
+		}
+		return &out, nil
+	}
+	return nil, lastErr
+}
+
+func (c *Client) ValidateToken(ctx context.Context, token string) (*TokenValidationResponse, error) {
```
jirevwe, line 168: The error values in this func have the same one as the func above

internal/pkg/sso/service/client.go
```diff
@@ -0,0 +1,264 @@
...
+		return &TokenValidationResponse{
+			Status:  out.Status,
+			Message: out.Message,
+			Data:    TokenValidationData{Payload: *p},
+		}, nil
+	}
+	return nil, lastErr
+}
+
+func (c *Client) GetAdminPortalURL(ctx context.Context, licenseKey, returnURL, successURL string) (*AdminPortalResponse, error) {
```
jirevwe, line 223: same as above

services/create_organisation.go
```diff
@@ -79,31 +80,83 @@ func (co *CreateOrganisationService) Run(ctx context.Context) (*datastore.Organi
...
+		BillingEmail: userEmail,
+		Host:         billingHost,
+	}
+	resp, createErr := billingClient.CreateOrganisation(ctx, orgData)
+	key := ""
+	if createErr == nil && resp != nil {
+		key = resp.Data.LicenseKey
+	}
+	if createErr != nil {
+		log.FromContext(ctx).WithError(createErr).Error("create_organisation: CreateOrganisation failed")
```
jirevwe: can we pass a service logger to this instead?

internal/pkg/sso/service/client.go
mekilis: The licenseKey is no longer passed to the API.

convoy-engineering:
@cursoragent review

PR #2563, opened 2026-02-12
fix(controlplane): default event list to 7 days and add events/index migration

mekilis:


mekilis:
Closing - changes may be covered elsewhere or no longer needed.

PR #2620, opened 2026-04-06
ci(release):  generate SPDX SBOM for published Docker images and upload as a workflow artifact

olamilekan000:
```
change adds an SBOM job to the image workflow after images are pushed.
SBOM is generated for each image and attached to the workflow artifacts. 
This makes it easier to trace exactly which packages and dependencies are included in a built image,
helping with vulnerability scanning, auditing, and supply chain verification.
```

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -176,3 +176,42 @@ jobs:
           docker buildx imagetools create -t ${{ env.IMAGE_NAME }}:main-${{ needs.set-image-tag.outputs.image_tag }} \
             ${{ env.IMAGE_NAME }}:main-${{ needs.set-image-tag.outputs.image_tag }}-amd64 \
             ${{ env.IMAGE_NAME }}:main-${{ needs.set-image-tag.outputs.image_tag }}-arm64
+
+  sbom:
+    name: SBOM Generation
+    runs-on: ubuntu-latest
+    needs: [build-and-push-default, set-image-tag]
+    permissions:
+      contents: read
+    steps:
+      - name: Login to Docker Hub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
+        with:
+          username: ${{ env.DOCKER_HUB_USERNAME }}
+          password: ${{ env.DOCKER_HUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-image.yml
```diff
@@ -176,3 +176,42 @@ jobs:
...
+          if [ "${{ needs.set-image-tag.outputs.release_kind }}" = "tag" ]; then
+            echo "ref=${{ env.IMAGE_NAME }}:${{ needs.set-image-tag.outputs.image_tag }}" >> "$GITHUB_OUTPUT"
+            echo "slug=${{ needs.set-image-tag.outputs.image_tag }}" >> "$GITHUB_OUTPUT"
+          else
+            echo "ref=${{ env.IMAGE_NAME }}:main-${{ needs.set-image-tag.outputs.image_tag }}" >> "$GITHUB_OUTPUT"
+            echo "slug=main-${{ needs.set-image-tag.outputs.image_tag }}" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Generate SPDX SBOM
+        uses: anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610
```
subomi: @olamilekan000, add a code comment for the action tag.

.github/workflows/build-image.yml
```diff
@@ -176,3 +176,42 @@ jobs:
...
+        with:
+          image: ${{ steps.image.outputs.ref }}
+          format: spdx-json
+          output-file: convoy.spdx.json
+          upload-artifact: false
+          upload-release-assets: false
+
+      - name: Upload SBOM artifact
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02
```
subomi: @olamilekan000, does this correctly attach to the uploaded Docker image?

.github/workflows/build-image.yml
olamilekan000: Like their version?

.github/workflows/build-image.yml
olamilekan000: No, this only includes it as part of the release artefact. Should we also include as part of the image?

PR #2626, opened 2026-04-17
refactor(docs): update doc generation

jirevwe:
## Summary                                                                                                                                                                                                        
                                                                                                                                                                                                                 
  - Replace all dashboard.getconvoy.io URLs across docs with region-specific {region}.getconvoy.cloud domains (US and EU)
  - Update openapi.yml with the latest auto-generated OpenAPI 3.0 spec from the convoy repo, including dual-server entries for US and EU regions                                                                 
  - Add 6 new API reference pages for endpoints that were missing from the docs: activate endpoint, test OAuth2 connection, import event types, refresh portal link token, validate source function, and test a  
  filter                                                                                                                                                                                                         
  - Update docs.json navigation to include the new API reference pages                                                                                                                                           
                                                                                                                                                                                                                 
  ## Related                                                         
                                                                                                                                                                                                                 
  - Companion PR: frain-dev/convoy-website#431                                                                                                                                       
  - Linear: [PDE-719](https://linear.app/convoy-engineering/issue/PDE-719/update-openapi-spec-base-urls)


docs/fix_openapi_spec.sh
```diff
@@ -0,0 +1,164 @@
...
+    fi
+
+    echo "Processing: $file"
+
+    # Create backup
+    cp "$file" "${file}.bak"
+
+    # Step 1: First fix nullable with sed (works on both JSON and YAML)
+    echo "  Step 1: Fixing nullable fields..."
+
```
mekilis: This rewrites Swagger 2.0 `x-nullable` to `nullable`, and the generated `docs/swagger.json` now fails schema validation (`.../properties/pagination must NOT have additional properties`). Please keep `x-nullable` in v2 and convert to `nullable` only for v3.

docs/docs.go
```diff
@@ -9997,7 +9997,12 @@ const docTemplate = `{
             "properties": {
                 "content": {},
                 "pagination": {
-                    "$ref": "#/definitions/datastore.PaginationData"
+                    "allOf": [
+                        {
+                            "$ref": "#/definitions/datastore.PaginationData"
+                        }
+                    ],
+                    "x-nullable": true
```
mekilis, line 10005: Please keep this as `x-nullable` for Swagger 2.0 output. Converting v2 nullable fields to `nullable` is what seems to be causing the regression in `docs/swagger.json`.

.githooks/pre-commit
```diff
@@ -15,9 +15,15 @@ then
       exit 0
 else
   # regenerate docs
-  swag init --generatedTime --parseDependency --parseInternal -g handlers/main.go -d api/ api/*
+  go run docs/annotate_dtos/main.go
+  swag init --generatedTime --parseDependency --parseDependencyLevel 3 --parseInternal -g handlers/main.go -d api/ api/*
   swag fmt -d ./api
-  go run v3gen/main.go
+  bash docs/fix_openapi_spec.sh
+  api-spec-converter --from=swagger_2 --to=openapi_3 -s yaml ./docs/swagger.yaml > ./docs/v3/openapi3.yaml
```
mekilis, line 27: This hook now requires `jq`, `yq`, and `api-spec-converter`. Please add `command -v` checks with install hints so failures are clear for non-`mise` setups.

.mise-tasks/gen/docs.sh
```diff
@@ -7,6 +7,21 @@
...
+#generate v2 openapi specs
 swag init --generatedTime --parseDependency --parseDependencyLevel 3 --parseInternal -g handlers/main.go -d api/ api/*
 swag fmt -d ./api
-go run v3gen/main.go
+
+# fix openapi2 specs (structural fixes, add proper produce/consume tags, replace x-nullable..)
+bash docs/fix_openapi_spec.sh
+
+#generate v3 specs
+api-spec-converter --from=swagger_2 --to=openapi_3 -s yaml ./docs/swagger.yaml > ./docs/v3/openapi3.yaml
```
mekilis, line 27: Looks good. Please add a small required-tools preflight (`jq`, `yq`, `api-spec-converter`) so it fails fast with a clear message.

docs/annotate_dtos/main.go
```diff
@@ -0,0 +1,208 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func main() {
+	dryRun := len(os.Args) > 1 && os.Args[1] == "--dry-run"
+
+	dtoDir := "./api/models"
+	if len(os.Args) > 2 {
```
mekilis, line 21: Custom directory parsing is off by one. `dtoDir` is read from `os.Args[2]`, so `go run docs/annotate_dtos/main.go ./api/models` does not work. Please parse with `flag` or accept `os.Args[1]` when it is not `--dry-run`.

mekilis (changes requested):

.githooks/pre-commit
```diff
@@ -14,10 +14,26 @@ if [ -z "$changed" ]
 then
       exit 0
 else
+  # check required tools
+  for tool in swag jq yq api-spec-converter openapi; do
```
mekilis, line 18: Partially fixed. Tool checks were added, but `openapi` is now required in preflight and still not pinned in `mise.toml`. Please add it so `mise install` matches the hook requirements.

.mise-tasks/gen/docs.sh
```diff
@@ -6,7 +6,33 @@
 
 set -e
 
+# preflight: check required tools
+for tool in swag jq yq api-spec-converter openapi; do
```
mekilis, line 10: Partially fixed. Preflight is in place, but it now checks `openapi` and `mise.toml` still does not include it.

docs/annotate_dtos/main.go
mekilis, line 21: Still open. Custom directory parsing still uses `os.Args[2]`, so `go run docs/annotate_dtos/main.go ./api/models` does not work.

mekilis (changes requested):

mekilis:
Also rebase from main. Thanks

.githooks/pre-commit
jirevwe: We should not require anyone to install these on their machines imo. That's why the dependency manager exists. These tools will never run in CI or in prod, it will always run locally.

.mise-tasks/gen/docs.sh
jirevwe: Same thing here. Forcing people to install a tool they'll only use to generate docs and making sure everyone is on the same version is a problem we struggled with for a long time and it makes no sense to keep supporting it

docs/annotate_dtos/main.go
jirevwe: It will never be used like that. We generate docs for the whole project, not per folder.

PR #2628, opened 2026-04-20
feat(controlplane): add logger lint rules

jirevwe:


.golangci.yml
```diff
@@ -69,6 +69,14 @@ linters:
     - dupl # Duplicate code
     - err113 # Error wrapping
   settings:
+    depguard:
+      rules:
+        deny-global-log:
```
mekilis, line 74: `deny-stdlib-log` is a better name I think

.golangci.yml
```diff
@@ -69,6 +69,14 @@ linters:
...
+          deny:
+            - pkg: "log"
+              desc: "Use pkg/logger instead of the standard log package"
+            - pkg: "log/slog"
+              desc: "Use pkg/logger instead of log/slog directly"
```
mekilis, line 79: Can we add an entry for the variants of `fmt.Print*`?

.golangci.yml
```diff
@@ -257,6 +267,15 @@ linters:
         linters:
           - all
           - cyclop
+      - path: "pkg/logger/"
```
mekilis, line 270: A nit pick, we are mixing quotes with and without for `- path` list items. Seems like a legacy issue.

mekilis (approved):

PR #2665, opened 2026-06-03
feat(dataplane,dashboard): configurable outbound request ID header per project

olamilekan000:
```
change adds per-project configuration for the outbound request ID header name (request_id_header), 
defaulting to X-Convoy-Idempotency-Key. 
Webhook deliveries send the event’s idempotency key under that header only 
(no duplicate legacy header when a custom name is set).
```
image

---

> [!NOTE]
> **Medium Risk**
> Changes webhook outbound headers and event-ingest validation on outgoing projects; misconfiguration could break subscribers expecting a stable id header, though defaults preserve prior behavior.
> 
> **Overview**
> Outgoing projects can set **`request_id_header`** so webhook deliveries put the event’s **`idempotency_key`** on that header name (default **`X-Convoy-Idempotency-Key`**), instead of always using the fixed legacy header. With a **non-default** header, publish paths **require** `idempotency_key` (API handlers, create-event services, dispatcher, and delivery workers fail closed without retries).
> 
> Project create/update **normalizes** the header, **rejects** custom values on **incoming** projects, and persists the field via a new **`project_configurations.request_id_header`** column. OpenAPI and the dashboard add outgoing-only configuration for this setting.
> 
> Also adds **`make ui-build`** / **`scripts/build-ui.sh`** and small dialog UX (Escape delegated on full-screen create/setup project modals).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8573f5998d41812a6a90941be4df61c64ee9c22. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


net/dispatcher.go
```diff
@@ -513,11 +513,11 @@ func (d *Dispatcher) sendWebhookInternal(ctx context.Context, endpoint string, j
 
 	header := httpheader.HTTPHeader(req.Header)
 	header.MergeHeaders(headers)
-	if len(idempotencyKey) > 0 && !hasHeader(header, "X-Convoy-Idempotency-Key") {
-		header["X-Convoy-Idempotency-Key"] = []string{idempotencyKey}
+	if len(idempotencyKey) > 0 && !util.IsStringEmpty(requestIDHeader) && !hasHeader(header, requestIDHeader) {
+		http.Header(header).Set(requestIDHeader, idempotencyKey)
```
mekilis: This feature exists to emit the customer's header verbatim. Their docs document it as `Split-Request-ID` with an example, but `.Set` canonicalizes the key so we ship `Split-Request-Id`, which doesn't match their contract. Setting it with `header[requestIDHeader] = []string{idempotencyKey}` preserves casing, but note the `canonicalizeHTTPHeader` call on line 520 would re-canonicalize it anyway, so this and that need fixing together. (HTTP/2 lowercases header names regardless, so consumers still need case-insensitive matching, but we shouldn't be the ones breaking the documented name.)

net/dispatcher.go
```diff
@@ -513,11 +513,11 @@ func (d *Dispatcher) sendWebhookInternal(ctx context.Context, endpoint string, j
...
 	}
 
-	req.Header = http.Header(header)
+	req.Header = canonicalizeHTTPHeader(http.Header(header))
```
mekilis: `canonicalizeHTTPHeader` runs over the whole map, so it also rewrites the casing of customer-supplied custom headers that used to pass through untouched. That's a silent change for every existing project, not just this feature, and breaks anyone relying on their own header casing. Can we scope it to just the request-id header?

net/dispatcher_test.go
```diff
@@ -895,10 +897,61 @@ func TestDispatcherSendRequest(t *testing.T) {
 	require.Equal(t, "custom-value", resp.RequestHeader.Get("X-Custom-Header"))
 }
 
+func TestDispatcherCustomRequestIDHeader(t *testing.T) {
+	const customHeader = "Split-Request-ID"
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		require.Equal(t, "stable-request-id", r.Header.Get(customHeader))
```
mekilis: This looks like it checks the custom header name, but it can't: the httptest server canonicalizes incoming keys and `Header.Get` canonicalizes the lookup, so `Split-Request-ID` and `Split-Request-Id` both pass. The casing issue in the dispatcher would stay green here. Assert on the raw request so the exact header name is actually covered.

worker/task/process_event_delivery.go
```diff
@@ -288,7 +288,20 @@ func ProcessEventDelivery(deps EventDeliveryProcessorDeps) func(context.Context,
...
+		resp, err := deps.Dispatcher.SendWebhookWithMTLS(
+			ctx,
+			targetURL,
+			sig.Payload,
+			project.Config.Signature.Header.String(),
+			header,
+			int64(cfg.MaxResponseSize),
+			eventDelivery.Headers,
+			project.Config.GetRequestIDHeader().String(),
+			eventDelivery.IdempotencyKey,
```
mekilis, line 371: Two problems. The header only goes out when the event has an idempotency key, so deliveries without one get no `Split-Request-ID` at all. And when it is sent, the value is that key (usually a ULID), not the UUID their contract specifies. As-is this doesn't meet the contract, which is a stable UUID on every delivery. The value and always-present pieces still need handling.

worker/task/process_meta_event.go
```diff
@@ -153,7 +153,19 @@ func sendUrlRequest(ctx context.Context, project *datastore.Project, metaEvent *
...
-	resp, err := dispatch.SendWebhook(ctx, url, sig.Payload, "X-Convoy-Signature", header, int64(cfg.MaxResponseSize), httpheader.HTTPHeader{}, dedup.GenerateChecksum(metaEvent.UID), httpDuration, "application/json")
+	resp, err := dispatch.SendWebhook(
+		ctx,
+		url,
+		sig.Payload,
+		"X-Convoy-Signature",
+		header,
+		int64(cfg.MaxResponseSize),
+		httpheader.HTTPHeader{},
+		project.Config.GetRequestIDHeader().String(),
```
mekilis: The project's outbound request-id header now applies to meta events too, with a checksum as the value. So a project set to `Split-Request-ID` stamps meta-event POSTs with `Split-Request-ID: <checksum>`. Meta events are a separate channel from webhook delivery, so reusing the customer-facing header name here looks out of scope. Intended?

debug.json
```diff
@@ -59,7 +59,7 @@
         "level": "info"
     },
     "host": "http://localhost:5011",
-    "root_path": "/testing",
+    "root_path": "",
```
mekilis: `root_path` flips from `/testing` to `""`, which looks like a local debug tweak that got swept in. Unrelated to the feature. Can we drop `debug.json` from the diff?

mekilis (commented):
Looks ok, but added a few comments. Also, please check the CI; one of the checks failed.

api/ui/build/index.html
```diff
@@ -1,388 +1,389 @@
-<!DOCTYPE html><html lang="en"><head>
+<!DOCTYPE html>
```
mekilis, line 1: `api/ui/build/` is checked in here, so the rebuilt `index.html` being in the diff is expected. Just confirm it's a clean regen from this PR's dashboard changes, not stale output or a hand-edit.

debug.json
mekilis: Still in the diff. Can we revert `debug.json` to match main?

worker/task/process_event_delivery.go
```diff
@@ -288,7 +288,20 @@ func ProcessEventDelivery(deps EventDeliveryProcessorDeps) func(context.Context,
...
-		resp, err := deps.Dispatcher.SendWebhookWithMTLS(ctx, targetURL, sig.Payload, project.Config.Signature.Header.String(), header, int64(cfg.MaxResponseSize), eventDelivery.Headers, eventDelivery.IdempotencyKey, httpDuration, contentType, mtlsCert)
+		resp, err := deps.Dispatcher.SendWebhookWithMTLS(
+			ctx,
+			targetURL,
+			sig.Payload,
+			project.Config.Signature.Header.String(),
+			header,
+			int64(cfg.MaxResponseSize),
+			eventDelivery.Headers,
+			project.Config.GetRequestIDHeader().String(),
```
mekilis, line 370: This works, thanks. Validating at publish with a 400 across the create paths, plus the dispatcher backstop, is the right call. One ask: add a code comment noting the producer must supply the UUID as the idempotency key, since that's the contract now.

mekilis (commented):
Added comments but please fix the failing tests.

mekilis:
Pushed a harden commit (`272552ce`):

- delivery + retry fail closed on missing idempotency key for custom request id headers (no retries)
- broadcast / fanout / dynamic validate before trial metering
- reject custom `request_id_header` on incoming create/update
- trim header names on read/normalize
- comment clarified: producer supplies any stable `idempotency_key`, not UUID-only

worker/task/process_event_delivery.go
mekilis: Done. Publish requires idempotency_key when a custom header is set (any stable value, not UUID-only). Dispatcher fails closed without retries if it is missing.

debug.json
mekilis: Done. debug.json matches main; not in the PR diff.

mekilis: Done. debug.json matches main; not in the PR diff.

worker/task/process_event_delivery.go
mekilis: Done. Comment is on the delivery path: producers must supply idempotency_key (any stable value) when a custom request id header is configured.

web/ui/dashboard/src/app/private/components/create-project-component/create-project-component.component.ts
```diff
@@ -459,6 +523,33 @@ export class CreateProjectComponent implements OnInit {
...
+
+	shouldShowConfigButton(config: { uid: string; show: boolean }, index: number): boolean {
+		if (config.show) {
+			return false;
+		}
+		if (this.isOutgoingOnlyConfig(config.uid)) {
+			return this.isOutgoingProject();
+		}
+		return index < 3;
+	}
```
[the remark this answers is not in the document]
mekilis: Fixed in 3c9877bf. Template now calls shouldShowConfigButton.

mekilis (approved):

PR #2669, opened 2026-06-17
fix(dataplane): fix batch replay oom

olamilekan000:
Change fixes a batch replay OOM risk where the handler defaulted to
a 2B(2000000000) page size and loaded all matching events into memory in one request.
It caps page size at 1000, paginates through results in
BatchReplayEventService, and applies pagination middleware to batch replay
routes.

---

> [!NOTE]
> **Medium Risk**
> Changes batch-replay pagination and HTTP error semantics (409 vs retryable service error), which can affect client retry behavior and how much work is enqueued per request.
> 
> **Overview**
> Fixes a batch-replay OOM path that defaulted to a ~2B page size and loaded every matching event in one query.
> 
> `BatchReplayEventService` now ignores caller page size/cursors, pages internally at **1000**, and fetches the next page *before* enqueueing the current one so a later fetch failure does not leave that page already queued (Redis replay IDs would re-fan-out on retry).
> 
> Incomplete runs that already enqueued work return **409** with success/failure counts; zero-success failures (ownership skips, fetch errors before any enqueue) stay on the retryable service-error path. Swagger documents the new 409.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 215ac3bd6dd40311a8478244786e8e9e61921ad6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


api/api.go
```diff
@@ -572,7 +571,7 @@ func (a *ApplicationHandler) mountControlPlaneRoutes(router chi.Router, handler
 							eventRouter.With(handler.RequireEnabledProject(), handler.RequireEnabledOrganisation()).Post("/fanout", handler.CreateEndpointFanoutEvent)
 							eventRouter.With(handler.RequireEnabledProject(), handler.RequireEnabledOrganisation()).Post("/broadcast", handler.CreateBroadcastEvent)
 							eventRouter.With(handler.RequireEnabledProject(), handler.RequireEnabledOrganisation()).Post("/dynamic", handler.CreateDynamicEvent)
-							eventRouter.With(handler.RequireEnabledProject(), handler.RequireEnabledOrganisation()).Post("/batchreplay", handler.BatchReplayEvents)
+							eventRouter.With(handler.RequireEnabledProject(), handler.RequireEnabledOrganisation(), middleware.Pagination).Post("/batchreplay", handler.BatchReplayEvents)
```
mekilis: This changes batch replay semantics when the caller includes list pagination params. The dashboard filter type already carries `next_page_cursor`, `prev_page_cursor`, and `direction`, and `batchReplayEvent()` sends the saved `queryParams` to `/events/batchreplay`. With this middleware, the replay starts from that cursor while `/countbatchreplayevents` still counts the full filter set, so the confirmation count can say N events but only the current page/window is replayed.

For a bulk replay endpoint, we should ignore caller cursors and use pagination only internally: start from the first cursor, force `direction=next`, cap `perPage`, and loop until done. Alternatively, do not attach `middleware.Pagination` to this route and let `BatchReplayEventService` build its own internal pageable.

PR #2699, opened 2026-06-28
fix(auth): return descriptive error when login fails with non-ServiceError

### LINKED ISSUE #2658: 🐛 LoginUser handler returns empty error message for non-ServiceError failures
## Description
The `LoginUser` handler in `api/handlers/auth.go` has a logic bug where the error message (`errMsg`) is initialized as an empty string and only populated for specific `ServiceError` codes. If the login error is not a `*services.ServiceError`, the response body will contain an empty error message.

## Context
- File: `api/handlers/auth.go:230-248`
- Component: Authentication / Login

## Current Behavior
```go
var errMsg string

if se, ok := err.(*services.ServiceError); ok {
    switch se.Code {
    case services.ErrCodeLicenseExpired:
        errMsg = se.ErrMsg
    default:
        errMsg = "Invalid credentials"
    }
}

_ = render.Render(w, r, util.NewErrorResponse(errMsg, http.StatusForbidden))
```

When `err` is not a `*services.ServiceError`, `errMsg` stays as `""` (empty string). The API returns:
```json
{"status": "error", "message": "", "code": 403}
```

This is problematic because:
1. Clients receive an unhelpful empty error message
2. It breaks API consistency — all other endpoints return descriptive error messages
3. It could confuse debugging efforts

## Expected Behavior
All authentication failures should return a descriptive error message, even if the error type is unexpected.

## Suggested Fix
Add a fallback for non-ServiceError cases:

```diff
  var errMsg string

  if se, ok := err.(*services.ServiceError); ok {
      switch se.Code {
      case services.ErrCodeLicenseExpired:
          errMsg = se.ErrMsg
      default:
          errMsg = "Invalid credentials"
      }
+ } else {
+     errMsg = "Authentication failed"
  }
```

## Impact
- **Severity: Low** — No security impact, but degrades API usability and debugging experience
- **Who is affected**: API clients integrating with Convoy authentication

_Positively — happy to submit a PR if this is welcome._

AruneshDwivedi:
LoginUser handler returned empty error message when the login error was not a *services.ServiceError, since errMsg was never set outside the type assertion. Adds fallback to "Authentication failed" for unexpected error types.

Fixes #2658

---

> [!NOTE]
> **Low Risk**
> Single-handler error-message fallback with no auth logic or credential-check changes.
> 
> **Overview**
> Fixes **empty error messages** on failed password login when `LoginUserService` returns an error that is not a `*services.ServiceError`.
> 
> In `LoginUser`, the handler now sets **`Authentication failed`** in an `else` branch after the existing `ServiceError` handling (license-expired message vs **Invalid credentials**). Clients always get a non-blank `403` message for unexpected failure types instead of an empty string.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d36bd83947e654af13c5a3402de1a1a4d57d6c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/api/handlers/auth.go
+++ b/api/handlers/auth.go
@@ -234,6 +234,8 @@ func (h *Handler) LoginUser(w http.ResponseWriter, r *http.Request) {
 			default:
 				errMsg = "Invalid credentials"
 			}
+		} else {
+			errMsg = "Authentication failed"
 		}
 
 		_ = render.Render(w, r, util.NewErrorResponse(errMsg, http.StatusForbidden))


mekilis:
Thanks for the patch, but closing this one. The fallback is harmless, yet the branch it adds isn't reachable today: every return path in `LoginUserService.Run` wraps the error in `*services.ServiceError`, so the new `else` never executes. This hardens against a hypothetical future change rather than fixing a live empty-message bug.

The more impactful issue is separate and untouched here: the `default` case collapses real internal failures (e.g. JWT generation errors) into "Invalid credentials". If we revisit this area, that's what we'd want to fix, with a test covering the non-`ServiceError` path.

PR #2706, opened 2026-07-07
fix(controlplane): redact secrets and close SSRF gaps across config, delivery, and notifications

mekilis:
## Summary

Batch of security hardening fixes across the control plane and delivery paths.

- **Config secret leak:** strip license/checkout secrets from `GetConfiguration` before the payload leaves the UI/portal route.
- **Token refresh gate:** gate token refresh on single-user-mode license, matching login. Fails closed on licenser error.
- **Response header redaction:** mask auth credentials in event-delivery and delivery-attempt responses. Partial last-4 reveal, short/low-entropy values fully masked, webhook signatures kept visible for debugging. Redaction is response-only via a shallow clone, so stored and dispatch values are untouched. Endpoint response headers (e.g. `Set-Cookie`) are redacted too.
- **SSRF: slack webhook:** validate `slack_webhook_url` on endpoint create/update and route slack notifications through the netjail dispatcher client.
- **SSRF: meta-event empty type:** validate the meta-event URL after normalizing an empty type to `http`, closing the empty-type bypass. Guard nil `SSL` on the project-update path (create defaults it, update did not).

## Test plan

- [ ] `go test ./api/... ./services/... ./internal/pkg/middleware/... ./util/...`
- [ ] Verify `GetConfiguration` response contains no license/checkout secrets
- [ ] Verify delivery-attempt / event-delivery responses mask auth headers but show `X-Convoy-Signature`
- [ ] Verify stored attempt/DB values remain unmasked
- [ ] Verify slack_webhook_url and meta-event URLs reject loopback/private/link-local targets
- [ ] Token refresh rejected in single-user mode without license

---

> [!NOTE]
> **High Risk**
> Touches authentication, license gating, secret exposure, and SSRF egress for user-controlled URLs—areas where mistakes have direct security impact, though behavior is largely fail-closed with tests.
> 
> **Overview**
> Security hardening across configuration exposure, auth, delivery APIs, and outbound HTTP.
> 
> **Configuration:** `GetConfiguration` now strips license keys, checkout nonces, and related billing secrets before responses reach UI/portal routes (billing still uses dedicated endpoints).
> 
> **Auth:** Login and token refresh share `IsPrimaryInstanceAdmin` so refresh cannot bypass single-user license limits. Transient license/DB failures return **5xx** (`ErrCodeInternal`) instead of “invalid credentials”; lapsed-license denials stay **403** on both paths.
> 
> **Delivery APIs:** Event delivery and delivery-attempt JSON redacts sensitive request/response headers for **portal-link** callers; JWT/API-key callers see raw values. Redaction is response-only (shallow clone); stored data and webhook **signatures** stay visible where intended.
> 
> **SSRF:** Slack webhook URLs are validated on endpoint create/update; meta-event URLs are validated after defaulting empty type to HTTP (with nil-safe `SSL`). Notifications use a dedicated HTTP client with connect-time private-IP blocking and enforced TLS (not webhook `insecure_skip_verify`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 146ea8eae544c3b6cc7a0c2824b35a577b0ddb0b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


api/models/delivery_attempt.go
```diff
@@ -0,0 +1,39 @@
...
+	if d.DeliveryAttempt == nil {
+		return []byte("null"), nil
+	}
+
+	clone := *d.DeliveryAttempt
+	clone.RequestHeader = m.RedactSensitiveHeaders(clone.RequestHeader)
+	clone.ResponseHeader = m.RedactSensitiveHeaders(clone.ResponseHeader)
+
+	return json.Marshal(&clone)
+}
```
jirevwe, line 43: Instead of redacting things in a disallow list, why not have an allow list of known safe headers then redact the rest. A customer can inject headers that are potentially sensitive and it will be shown on the UI

mekilis: This has been refactored in scope to portal links alone. But I can flip that to an allow list of known-safe headers if we want portal to fail closed on customer-injected header names.

PR #2777, opened 2026-08-10
docs(docs): correct advanced signatures defaults

ScMofeoluwa:
## Summary

The `advanced_signatures` field is documented as defaulting to `false` when omitted, but its behavior differs between endpoint creation and updates.

For endpoint creation, `services/create_endpoint.go` defaults the field to `true` when omitted. For endpoint updates, `services/update_endpoint.go` only changes the field when a value is provided, so omission retains the current value.

This PR corrects both model comments and regenerates the OpenAPI documentation to reflect the existing behavior. It does not change application behavior.

## Testing

- Ran `make generate_docs`
- Swagger 2.0 JSON validation passed
- Swagger 2.0 YAML validation passed
- OpenAPI 3 validation passed
- Ran `git diff --check`

---

> [!NOTE]
> **Low Risk**
> Documentation-only change with no application behavior impact.
> 
> **Overview**
> Corrects outdated docs for the `advanced_signatures` field on create/update endpoint models.
> 
> Previously documented as defaulting to `false` when omitted. Docs now state that create defaults to **true** for outgoing projects (incoming always uses advanced), and update omits leave the current value unchanged. Regenerates OpenAPI/Swagger artifacts to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 871eacee06384bea0c26cd8f253ebae0c5f24ab5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mekilis:
Thanks for catching this, the old text was wrong.

The fix was still incomplete though: the default is project-type dependent. `services/create_endpoint.go` forces advanced signatures for incoming projects regardless of what the client sends, and only applies the `true` default to outgoing ones. `services/update_endpoint.go` ignores the field entirely unless the project is outgoing. So a single unconditional default is still misleading for half the cases.

I pushed 871eacee on top of yours to scope both descriptions:

- Create: "Only applies to outgoing projects, where it defaults to true when omitted. Incoming projects always use advanced signatures."
- Update: "Only applies to outgoing projects; omit it to keep the current value. Incoming projects always use advanced signatures."

Kept each to a single comment line on purpose. swag turns every line break into a paragraph break in the generated spec, and a wrapped two-line comment split the sentence mid-clause in `docs/v3/openapi3.yaml`. Regenerated all six artifacts with `make generate_docs`.

For context, the original "we default to false" was not invented. That is the `2024-01-01` behavior, where `CreateEndpointMigration` sets the field to false when it is omitted. It only applies to clients pinning that version, and the spec documents the current default (`2025-11-24`), so `true` is correct here.

Two CI notes: the title check needs another maintainer approval to re-run after your rename, I will approve it. The failing unit shard is a testcontainers Docker Hub pull timeout, unrelated to this change.

ScMofeoluwa:
Thanks for catching that and for the extra context. I had noticed the incoming/outgoing difference in the implementation, but the version-specific behavior wasn’t on my radar. The scoped wording makes the docs much clearer. Appreciate you cleaning that up and rerunning the checks

mekilis (approved):
Approved. The wording now matches create/update behavior for both project types, and the regenerated spec is clean.

Flagging for the record that the head commit is mine, so this approval covers your original commit rather than being a fully independent review of the branch. Docs-only, so I am comfortable with it.

PR #2780, opened 2026-08-12
feat(dataplane): preserve endpoint URLs on deliveries

ScMofeoluwa:
## Summary

When an endpoint URL is updated, existing event deliveries retain only the endpoint ID. The dashboard then resolves the URL from the endpoint’s current configuration, making it impossible to determine which URL was associated with an older delivery.

This change:

- Stores the endpoint URL on each delivery when it is created.
- Exposes the value as `endpoint_url` in delivery responses.
- Displays the historical URL in the dashboard.
- Keeps `target_url` unchanged for dynamic deliveries.
- Preserves existing retry behavior, which continues using the current endpoint URL for static deliveries.
- Shows `-` for older deliveries that predate this field and have no historical snapshot.

## Testing

- OpenAPI generation and validation passed.
- Migration validation passed.
- Resolver test passed.
- Event-delivery package compilation passed.
- `git diff --check` passed.

Full integration tests were not runnable locally because the test environment requires a JWT secret

---

> [!NOTE]
> **Medium Risk**
> Touches core `event_deliveries` schema and migration paths (including table-rebuild helpers in impl), but delivery HTTP behavior is explicitly unchanged; main risk is migration/deploy on large partitioned tables and API consumers seeing a new optional field.
> 
> **Overview**
> Adds **`endpoint_url`** on event deliveries so the dashboard and API can show the endpoint URL at creation time instead of always reflecting the live endpoint config.
> 
> On delivery creation, the worker snapshots the subscription endpoint’s URL into **`EndpointURL`**. Persistence is wired through the model, SQL/sqlc queries, batch create, and a new migration (`endpoint_url` TEXT). OpenAPI/Swagger specs expose the field on delivery responses.
> 
> The **dashboard** event delivery details view prefers **`target_url`**, then **`endpoint_url`**, for the copyable delivery URL (no longer falling back to **`endpoint_metadata.url`** for that display). **`target_url`** stays the source for dynamic/templated deliveries.
> 
> **Retry routing is unchanged**: dispatch still resolves URLs from **`target_url`** / the current endpoint URL; stored **`endpoint_url`** is not used for HTTP delivery (covered by a new resolver test).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 333b0e2aac9e49b7d34aeee9cb68df1a99ab7e4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


worker/task/process_event_creation.go
```diff
@@ -357,6 +357,11 @@ func writeEventDeliveriesToQueue(ctx context.Context, opts WriteEventDeliveriesT
 			deliveryStatus = datastore.DiscardedEventStatus
 		}
 
+		var endpointURL string
+		if s.Endpoint != nil {
+			endpointURL = s.Endpoint.Url
+		}
```
mekilis, line 363: This stores the raw `endpoint.Url`, so a templated endpoint snapshots the template (`/orders/{reference}/callback`) instead of the URL that was sent. Please read the concrete value off the attempt.

web/ui/dashboard/src/app/private/pages/project/events/event-delivery-details/event-delivery-details.component.html
```diff
@@ -219,14 +219,14 @@ <h3 class="font-heading font-medium text-[16px] leading-[24px] text-new.text-pri
         }
       </div>
 
-      <convoy-copy-button [text]="eventDelsDetails?.target_url || eventDelsDetails?.endpoint_metadata?.url || ''" show-icon="false" notificationText="Endpoint URL copied to clipboard">
+      <convoy-copy-button [text]="eventDelsDetails?.target_url || eventDelsDetails?.endpoint_url || ''" show-icon="false" notificationText="Endpoint URL copied to clipboard">
         <span class="relative inline-flex items-center max-w-[370px] group cursor-pointer">
-          <span class="truncate font-body text-[13px] text-new.text-secondary">{{ eventDelsDetails?.target_url || eventDelsDetails?.endpoint_metadata?.url || '-' }}</span>
+          <span class="truncate font-body text-[13px] text-new.text-secondary">{{ eventDelsDetails?.target_url || eventDelsDetails?.endpoint_url || '-' }}</span>
```
mekilis, line 224: This drops the live URL fallback, so every delivery created before the migration shows `-` here, which is all existing data on an upgrade. Please keep `endpoint_metadata?.url` last in the chain.

mekilis (commented):
Thanks for the quick turnaround on this.

This does exactly what we described in Slack. That guidance was wrong, so apologies for the rework.

The URL that was actually hit is already stored on `delivery_attempts.url`. `endpoint_url` gives `event_deliveries` a second `*_url` that means nearly the same thing as `target_url`, and it costs a column on the deliveries table plus the partition rewrite SQL, for a field the UI only reads when `target_url` is empty.

Please drop the column and read the history off the attempt instead. A delivery with no attempt yet has nothing to show, and the live endpoint URL is right there anyway since that is where it will go.

On the tests, you only need `CONVOY_JWT_SECRET` and `CONVOY_JWT_REFRESH_SECRET`. CI sets them to `test-access-secret` and `test-refresh-secret` in `.github/workflows/integration-tests.yml`, and `mise.toml` already has dev values, so `mise exec -- go test ./...` picks them up. Postgres on 5432 (`postgres`/`postgres`/`convoy`) and Redis on 6379 match the workflow services.

Thanks again for picking this up.

ScMofeoluwa:
> Thanks for the quick turnaround on this.
> 
> This does exactly what we described in Slack. That guidance was wrong, so apologies for the rework.
> 
> The URL that was actually hit is already stored on `delivery_attempts.url`. `endpoint_url` gives `event_deliveries` a second `*_url` that means nearly the same thing as `target_url`, and it costs a column on the deliveries table plus the partition rewrite SQL, for a field the UI only reads when `target_url` is empty.
> 
> Please drop the column and read the history off the attempt instead. A delivery with no attempt yet has nothing to show, and the live endpoint URL is right there anyway since that is where it will go.
> 
> On the tests, you only need `CONVOY_JWT_SECRET` and `CONVOY_JWT_REFRESH_SECRET`. CI sets them to `test-access-secret` and `test-refresh-secret` in `.github/workflows/integration-tests.yml`, and `mise.toml` already has dev values, so `mise exec -- go test ./...` picks them up. Postgres on 5432 (`postgres`/`postgres`/`convoy`) and Redis on 6379 match the workflow services.
> 
> Thanks again for picking this up.

Thanks, this is a fair point. I was aware that `delivery_attempts.url` contains the concrete URL that was called, but my concern is specifically at the delivery level.

After an endpoint URL changes, an older delivery can still be returned with the endpoint’s current URL. Consumers then have to fetch the attempts, or correlate the delivery against endpoint history, to determine which endpoint URL that delivery belonged to.

For context, I currently have to maintain a separate table on my side to track endpoint URL changes so I can associate historical deliveries with the URL configured at that point in time. A delivery-level snapshot in Convoy would remove that duplication and let consumers rely on the delivery API directly.

I’m looking for the delivery API itself to expose that historical value directly, for example:

```
{
  uid: delivery_123,
  endpoint_url: https://old.example.com/webhook
}
```

I intentionally kept `endpoint_url` separate from `target_url`: `endpoint_url` would be historical delivery metadata, while `target_url` retains its existing routing semantics. For templated endpoints, I agree that `delivery_attempts.url` remains the source of the concrete URL that was actually called, but that does not replace the delivery-level value.

Would Convoy consider retaining this delivery-level snapshot, or exposing an equivalent delivery-level historical URL through the API? That would avoid requiring every consumer to make a second request and reconstruct the relationship themselves.

mekilis:
Thanks, the use case is clearer now and you are right that there is a gap. I checked: `target_url` is only set on the dynamic path when the URL differs from the endpoint, so for ordinary deliveries it is empty and the delivery comes back with the endpoint's current URL. My review called that the rare case when it is the common one.

The stored column is still not the path, for three reasons:

- Non-templated endpoint: `delivery_attempts.url` is the endpoint URL as dialed, so the column duplicates it.
- Templated endpoint: the column stores the template, which is not the URL used. You agreed the attempt wins there.
- Existing data: attempts have carried this for years. A new column is NULL for every delivery you already have, which leaves your original problem in place.

The only case attempts miss is a delivery that was never attempted, and there the live endpoint URL is correct, since that is where it will go.

On your side table, you can drop it now rather than waiting on this PR. `/eventdeliveries/{id}/deliveryattempts` returns the URL each attempt actually used, on the deliveries you already have, so you do not need to track endpoint history yourself.

Before settling on what the API should expose, one question, because it changes the answer. Is your consumer per-delivery or list-driven? If you fetch one delivery at a time, the attempts call is a single extra request and folding the URL into the delivery response is a small change. If you are walking the deliveries list and want the URL on every row, that is the harder problem and worth designing properly: `delivery_attempts` is partitioned on `(project_id, created_at)`, so a per-delivery lookup prunes by project but not by time and probes every partition from the delivery forward, once per row. Which one you need decides whether this is a response field, a join, or something else.

Please hold off reworking the PR until we agree on that shape. You have rebuilt this once already and I would rather not have you do it twice.

Thanks for the detail on the use case, it changed part of my read here.

mekilis:
One clarification on my last comment, so you are not calibrating on the wrong thing.

Convoy's own dashboard does not render this in either shape. The deliveries list has no URL column, and the details page reads `target_url` then `endpoint_metadata.url`. So there is no internal driver pushing this into the schema, and a list-shaped answer is more likely to end at "use the attempts endpoint" than at a stored column. I would rather say that now than have you read the question as leaning the other way.

Your report did find something real in Convoy itself though. The URL row on the delivery details page falls back to the endpoint's current URL, so after a URL change it shows a URL that delivery may never have gone to, while the attempts tab beside it holds the truth. Nobody had raised that. It is misleading and worth fixing on our side, and the fix is the same attempt-sourced read, independent of whatever we decide about the API.

Still interested in the use case either way.

ScMofeoluwa:
> Thanks, the use case is clearer now and you are right that there is a gap. I checked: `target_url` is only set on the dynamic path when the URL differs from the endpoint, so for ordinary deliveries it is empty and the delivery comes back with the endpoint's current URL. My review called that the rare case when it is the common one.
> 
> The stored column is still not the path, for three reasons:
> 
> * Non-templated endpoint: `delivery_attempts.url` is the endpoint URL as dialed, so the column duplicates it.
> * Templated endpoint: the column stores the template, which is not the URL used. You agreed the attempt wins there.
> * Existing data: attempts have carried this for years. A new column is NULL for every delivery you already have, which leaves your original problem in place.
> 
> The only case attempts miss is a delivery that was never attempted, and there the live endpoint URL is correct, since that is where it will go.
> 
> On your side table, you can drop it now rather than waiting on this PR. `/eventdeliveries/{id}/deliveryattempts` returns the URL each attempt actually used, on the deliveries you already have, so you do not need to track endpoint history yourself.
> 
> Before settling on what the API should expose, one question, because it changes the answer. Is your consumer per-delivery or list-driven? If you fetch one delivery at a time, the attempts call is a single extra request and folding the URL into the delivery response is a small change. If you are walking the deliveries list and want the URL on every row, that is the harder problem and worth designing properly: `delivery_attempts` is partitioned on `(project_id, created_at)`, so a per-delivery lookup prunes by project but not by time and probes every partition from the delivery forward, once per row. Which one you need decides whether this is a response field, a join, or something else.
> 
> Please hold off reworking the PR until we agree on that shape. You have rebuilt this once already and I would rather not have you do it twice.
> 
> Thanks for the detail on the use case, it changed part of my read here.

My use case is list-driven. I need to page through delivery history and have the URL available for each delivery, rather than fetching one delivery and then making a separate attempts request. The concern is not only the dashboard details page; it is consumers needing to reconcile a delivery list after the endpoint URL changes. An N+1 attempts lookup would avoid the new column but would make the list API expensive and require every consumer to reproduce that logic

mekilis:
Thanks, that answers it.

You are right that consumers should not fetch attempts per row. If we put the URL on the list it would be a join, not an N+1. For now, we are still not going to add a stored column. Please use `/eventdeliveries/{id}/deliveryattempts` for the historical URL; that already covers the data you have. We will fix the dashboard details row later.

No need to rework the PR. Thanks again for raising this.

ScMofeoluwa:
> Thanks, that answers it.
> 
> You are right that consumers should not fetch attempts per row. If we put the URL on the list it would be a join, not an N+1. For now, we are still not going to add a stored column. Please use `/eventdeliveries/{id}/deliveryattempts` for the historical URL; that already covers the data you have. We will fix the dashboard details row later.
> 
> No need to rework the PR. Thanks again for raising this.

Thanks for taking the time to work through this with me. The distinction between delivery-level history and attempt-level URLs is clear now, and I agree that adding endpoint_url is not the right change for this PR.

I’ll close this PR as requested. The review clarified that the dashboard’s current URL fallback is misleading, and that a future fix for list-driven consumers should expose the relevant attempt URL through a join in the delivery response, rather than adding a new delivery column or making per-row requests.

Thanks again for the thoughtful review.

PR #2798, opened 2026-08-19
fix(infra): run local installer from published image

motunrayokoyejo:
Reworks `install-local.sh` into a fully self-contained quick-start installer that no longer clones the repo or contacts GitHub at all: it generates the local stack config files directly and runs Convoy from the prebuilt `getconvoy/convoy:latest` Docker image

---

> [!NOTE]
> **Low Risk**
> Changes are limited to the local installer script and generated dev configs; no production app or auth runtime code paths are modified.
> 
> **Overview**
> **Reworks `install-local.sh` into a self-contained quick-start** that no longer clones the Convoy repo or requires Git. Install flow is **`generate_config_files`** instead of **`prepare_repo`**, writing `docker-compose.yml`, Caddyfile, PgBouncer env, and default `convoy.json` under `$INSTALL_DIR/configs/local`.
> 
> Services use **`CONVOY_IMAGE`** (default `getconvoy/convoy:latest`) for migrate/web/agent. Existing **`convoy.json` is left untouched** on re-runs so local tweaks persist; **`ensure_local_config`** no longer restores files from git.
> 
> Prerequisites drop the Git check. Success output adds the **default dashboard login** (`superuser@default.com` / `default`) alongside the URL.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e48e64de796f426942070d5288a2c64caac8706. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


install-local.sh
```diff
@@ -251,42 +247,233 @@ write_compose_file() {
...
+        }
+    }
+}
+EOF
   else
-    if [ -d "$INSTALL_DIR" ] && [ "$(ls -A "$INSTALL_DIR" 2>/dev/null)" ]; then
-      die "Install directory '$INSTALL_DIR' exists and is not a git repo. Use an empty path or set CONVOY_INSTALL_DIR to another directory."
-    fi
-    git clone "$REPO_URL" "$INSTALL_DIR"
+    log "Existing convoy.json found; keeping it"
```
mekilis: Keeping `convoy.json` on re-run is right for host/auth tweaks. It is wrong for the redis block: the compose you just wrote no longer mounts a CA or serves TLS. Patch redis to plain `redis://` whenever this function rewrites compose, including when the json file already exists.

install-local.sh
```diff
@@ -251,42 +247,233 @@ write_compose_file() {
...
+        volumes:
+            - ./conf/:/bitnami/pgbouncer/conf/
+            - ./conf/userlists.txt:/bitnami/userlists.txt
+
+    redis_server:
+        image: redis:7-alpine
+        restart: unless-stopped
+        command: ["redis-server", "--maxmemory", "256mb", "--maxmemory-policy", "allkeys-lru"]
+        volumes:
+            - redis_data:/data
```
mekilis, line 361: This is the other half of the same break. A first-run json matches it. An old cloned json does not. The migrate has to happen in the keep-json path, not by restoring TLS here.

mekilis (commented):
Please fix the reinstall Redis mismatch before merge. Title also needs a conventional rename.

Re-running this installer keeps an existing `convoy.json` (including `rediss` and `tls_ca_cert_file` from the old cloned stack) while always rewriting Compose so Redis has no TLS and the CA cert is not mounted. Startup pings Redis, so web/agent never become healthy and the wait on `/healthz` times out. That is the previous default `$HOME/convoy` install.

When keeping `convoy.json`, rewrite the redis section to match the stack this script just wrote (`scheme: redis`, no `tls_ca_cert_file`). Do not treat redis transport as a customization the compose rewrite is allowed to invalidate.

Also rename the PR to `fix(infra): run local installer from published image`. `script` is not an allowed scope.

Docker compose smoke still runs `configs/local/smoke-test.sh` (TLS Redis), not this script, so CI will not catch the break.

motunrayokoyejo:
> Please fix the reinstall Redis mismatch before merge. Title also needs a conventional rename.
> 
> Re-running this installer keeps an existing `convoy.json` (including `rediss` and `tls_ca_cert_file` from the old cloned stack) while always rewriting Compose so Redis has no TLS and the CA cert is not mounted. Startup pings Redis, so web/agent never become healthy and the wait on `/healthz` times out. That is the previous default `$HOME/convoy` install.
> 
> When keeping `convoy.json`, rewrite the redis section to match the stack this script just wrote (`scheme: redis`, no `tls_ca_cert_file`). Do not treat redis transport as a customization the compose rewrite is allowed to invalidate.
> 
> Also rename the PR to `fix(infra): run local installer from published image`. `script` is not an allowed scope.
> 
> Docker compose smoke still runs `configs/local/smoke-test.sh` (TLS Redis), not this script, so CI will not catch the break.

resolved

